PR #11, opened 2022-05-24
Add script to run CI on changed RBI files only

adisonlampert:
Added a script that populates the next command with modified/new RBIs so that the CI would only run on those modified/new RBIs.
Examples:
- #13 Adds a new RBI successfully, only linting & typechecking the new file
- #14 Fails typechecking on a modified RBI, only linting & typechecking the modified file

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,20 +2,25 @@ name: CI
 
 on: [push, pull_request]
 
+env:
+  GIT_DEFAULT_BRANCH: ${{github.event.repository.default_branch}}
+
 jobs:
   build:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
-    name: CI
     steps:
-      - uses: actions/checkout@v2
+      - name: Checkout repo
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
       - name: Set up Ruby
         uses: ruby/setup-ruby@v1
         with:
           ruby-version: '2.7'
           bundler-cache: true
       - name: Lint RBI files
-        run: bundle exec rubocop
+        run: scripts/run_on_changed_rbis "bundle exec rubocop"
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -2,20 +2,25 @@ name: CI
...
         uses: ruby/setup-ruby@v1
         with:
           ruby-version: '2.7'
           bundler-cache: true
       - name: Lint RBI files
-        run: bundle exec rubocop
+        run: scripts/run_on_changed_rbis "bundle exec rubocop"
       - name: Typecheck RBI files
-        run: scripts/check_types
+        run: scripts/run_on_changed_rbis scripts/check_types
```
RyanBrushett, line 26: This looks wrong. I'm assuming it should be something like this:

```suggestion
        run: scripts/run_on_changed_rbis scripts "./check_types"
```

scripts/check_types
```diff
@@ -6,6 +6,7 @@ require "open3"
 success = true
 files_with_errors = []
 
+*files = ARGV
 files = Dir.glob("./rbi/**/*").sort if ARGV.empty?
```
RyanBrushett, line 10: Maybe better to change this check to be `if files.empty?`

Or better yet, refactor this couple of lines to be something like

```rb
files = ARGV.empty? ? Dir.glob.etc. : ARGV
```

or it's expanded form

```rb
files = if ARGV.empty?
  # blah blah
else
  ARGV
end
```

RyanBrushett (changes requested):

.github/workflows/ci.yml
RyanBrushett: Ignore this, I misread

RyanBrushett (approved):
OK yeah rad, ignore my ❌ , I misread something. One formatting ⛏️ but otherwise this looks neat to me.

PR #38, opened 2022-06-09
Add a pre-push git hook

adisonlampert:
Added a `pre-push` git hook with the main intention of preventing people from pushing annotations for private gems. Included the other CI tests as well for fun.

Usually hooks go in `.git/hooks` but `.git` is not tracked so I looked into solutions for how to share git hooks and found that by placing the git hooks in an arbitrary folder (I named this one `.githooks`) it could be tracked and included with the repository and then anyone using the repository could run `git config core.hooksPath .githooks` to configure it.

Making a branch with an example of a failure would be difficult since the `pre-push` git hook prevents me from pushing it but you can use any of the failure examples from the past to test with and they will still all fail, just this time before you even push:
- #3 
- #4 
- #19 
- #20 
- #35

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -1,0 +1,47 @@
+#! /usr/bin/env ruby
+
+$stderr.puts("Running pre-push validation...")
+
+lines = `git fetch origin main 2> /dev/null && git diff --name-only origin/main | grep -E "\.rbi$|index.json"`.lines
+files = lines.map(&:strip).select { |file| File.file?(file) }
+rbis = files.select { |file| File.extname(file) === ".rbi"}
+
+if lines.empty?
+  $stderr.puts("Nothing to check")
+  $stderr.puts("\n")
+  exit(0)
+end
+
+success = true
+
+unless rbis.empty?
+  $stderr.puts("Linting RBI files...")
+  unless system("bundle exec rubocop #{rbis.join(" ")}")
+    success = false
+  end
+  $stderr.puts("Type checking RBI files...")
+  unless system("scripts/check_types #{rbis.join(" ")}")
+    success = false
+  end
+  $stderr.puts("Checking that new RBI files belong to public gems...")
+  unless system("scripts/check_gems_are_public #{rbis.join(" ")}")
+    success = false
+  end
+end
+
+$stderr.puts("Linting index.json...")
+unless system("scripts/lint_index_json")
+  success = false
+end
+$stderr.puts("Checking that index.json entries match rbi/annotations/*.rbi files...")
+unless system("scripts/match_index_entries_with_rbis")
+  success = false
+end
+
+unless success
+  $stderr.puts("Cancelling the push as the code fails at least one check (see above)")
+  exit(1)
+end
+
+$stderr.puts("\n")
+exit(0)
--- a/README.md
+++ b/README.md
@@ -34,6 +34,14 @@ The CI for this repository is written using GitHub Actions. Here is a list of cu
 
 All scripts used in the CI checks are located in the `scripts` folder.
 
+To avoid pushing a PR with errors, configure the `pre-push` git hook by running:
+
+```shell
+$ git config core.hooksPath .githooks
+```
+
+This git hook will run most of the CI checks and block the push if you don't pass all the tests.
+
 ## Contributing
 
 Bug reports and pull requests are welcome on GitHub at https://github.com/Shopify/rbi-central.
--- a/scripts/check_gems_are_public
+++ b/scripts/check_gems_are_public
@@ -18,7 +18,7 @@ files.each do |file|
 
   if version === "9001.0" || version === "unknown"
     $stderr.puts("  Error: `#{gem}` doesn't seem to be a public")
-    $strerr.puts("  Make sure your gem is available at https://rubygems.org/gems/#{gem}")
+    $stderr.puts("  Make sure your gem is available at https://rubygems.org/gems/#{gem}")
     success = false
   end
 end


.githooks/pre-push
```diff
@@ -0,0 +1,27 @@
+#! /usr/bin/env ruby
+
+puts "Attempting to push..."
+
+lines = `git fetch origin main && git diff --name-only origin/main | grep "\.rbi$"`.lines
```
Morriar, line 5: This won't work if there is a change in `index.json` only.

.githooks/pre-push
```diff
@@ -0,0 +1,27 @@
+#! /usr/bin/env ruby
+
+puts "Attempting to push..."
```
Morriar, line 3: I was surprised by the output in my command line.

Maybe something more obvious:
```suggestion
puts "Running pre-push validation..."
```

README.md
```diff
@@ -34,6 +34,8 @@ The CI for this repository is written using GitHub Actions. Here is a list of cu
 
 All scripts used in the CI checks are located in the `scripts` folder.
 
+To avoid accidently pushing information about your private gem, configure the `pre-push` git hook by running `git config core.hooksPath .githooks` in your terminal. This git hook will run most of the CI checks and block the push if you don't pass all the tests.
```
Morriar, line 37: it's not just about private things. It's also all kind of errors.

Let's also make the command more visible so we found it easily when scorlling:

~~~md
```shell
$ git config core.hooksPath .githooks
```
~~~

.githooks/pre-push
Morriar: For example I pushed this and the hooks didn't catch anything: https://github.com/Shopify/rbi-central/compare/at-test?expand=1

README.md
```diff
@@ -34,7 +34,11 @@ The CI for this repository is written using GitHub Actions. Here is a list of cu
 
 All scripts used in the CI checks are located in the `scripts` folder.
 
-To avoid accidently pushing information about your private gem, configure the `pre-push` git hook by running `git config core.hooksPath .githooks` in your terminal. This git hook will run most of the CI checks and block the push if you don't pass all the tests.
+To avoid pushing a PR with errors, configure the `pre-push` git hook by running:
+```shell
+$ git config core.hooksPath .githooks
+```
+This git hook will run most of the CI checks and block the push if you don't pass all the tests.
```
Morriar, line 41:
~~~suggestion
To avoid pushing a PR with errors, configure the `pre-push` git hook by running:

```shell
$ git config core.hooksPath .githooks
```

This git hook will run most of the CI checks and block the push if you don't pass all the tests.
~~~

.githooks/pre-push
```diff
@@ -12,15 +12,25 @@ if lines.empty?
...
 puts "4. Lint index.json"
-system("scripts/lint_index_json") ? continue : exit(1)
+unless system("scripts/lint_index_json") 
+  exit(1)
+end
 puts "5. Check index.json entries match rbi/annotations/*.rbi files"
-system("scripts/match_index_entries_with_rbis") ? continue : exit(1)
+unless system("scripts/match_index_entries_with_rbis")
+  exit(1)
+end
```
Morriar, line 33: Can we run all the tests rather than exiting early. So if there is a problem with a a RBI _and_ a lint error in the `index.json` file I won't need to push twice to see it?

adisonlampert: Okay! Would the best way to approach this to be some sort of variable like `success = true` and set it to false if any of the tests fail in place of `exit(1)` so for example:
```
unless system("scripts/match_index_entries_with_rbis")
  success = false
end
```
and then at the end having logic to `exit(0)` if success and `exit(1)` otherwise?

.githooks/pre-push
```diff
@@ -0,0 +1,45 @@
...
+  success = false
+end
+puts "5. Check index.json entries match rbi/annotations/*.rbi files"
+unless system("scripts/match_index_entries_with_rbis")
+  success = false
+end
+
+if success
+  puts "Everything looks good. Pushing to remote..."
+  puts "\n"
```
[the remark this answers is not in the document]
adisonlampert: That was just to make it a little more readable since right after that it goes into all the git stuff about the push.

.githooks/pre-push
```diff
@@ -0,0 +1,45 @@
+#! /usr/bin/env ruby
+
+puts "Running pre-push validation..."
+
+lines = `git fetch origin main && git diff --name-only origin/main | grep -E "\.rbi$|index.json"`.lines
+files = lines.map(&:strip).select { |file| File.file?(file) }
+
+if lines.empty?
+  puts "Nothing to check"
+  puts "\n"
```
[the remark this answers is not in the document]
Morriar: I think we can redirect this output to /dev/null.

See how git is using `stderr` for these diagnostics? We can leverage this by redirecting stderr: `git fetch origin main 2> /dev/null && git diff --name-only origin/main` and not pollute the output.

.githooks/pre-push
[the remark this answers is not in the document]
Morriar: Ok, let's keep this one then:

```rb
$stderr.puts "\n"
exit(0)
```

Morriar (approved):

.githooks/pre-push
```diff
@@ -0,0 +1,43 @@
...
+  success = false
+end
+$stderr.puts("Linting index.json...")
+unless system("scripts/lint_index_json")
+  success = false
+end
+$stderr.puts("Checking that index.json entries match rbi/annotations/*.rbi files...")
+unless system("scripts/match_index_entries_with_rbis")
+  success = false
+end
```
Morriar, line 35: I added a `"test": {}` entry in the index file and somehow this didn't catch any problem.

.githooks/pre-push
```diff
@@ -0,0 +1,43 @@
...
+$stderr.puts("Type checking RBI files...")
+unless system("scripts/check_types #{files.join(" ")}")
+  success = false
+end
+$stderr.puts("Checking that new RBI files belong to public gems...")
+unless system("scripts/check_gems_are_public #{files.join(" ")}")
+  success = false
+end
```
Morriar, line 27: I only changed the `index.json` file and now we're trying to validate it against Rubygems:

```
Checking that new RBI files belong to public gems...
Checking if gem `index` is public...

All gems are public, good job!
```

Morriar (commented):
I found a few issues but it prevented the push.

.githooks/pre-push
Morriar: Ok, it works, not sure why I didn't get the error the first time 👍

Morriar (approved):

.githooks/pre-push
RyanBrushett: It successfully stopped push for me as well 👍

RyanBrushett (changes requested):
It works but there's [a typo on line 21 of the check gems are public script](https://github.com/Shopify/rbi-central/blob/main/scripts/check_gems_are_public#L21) so I'm not sure if it works for the right reasons hahah. Can you correct that typo and re-test as part of this PR pls?

RyanBrushett (approved):
Worked like a charm. Thanks for this!

PR #39, opened 2022-06-14
Add a Sorbet config so we can use the VSCode plugin

Morriar:
Supressing error codes [5002](https://sorbet.org/docs/error-reference#5002), [5020](https://sorbet.org/docs/error-reference#5020), [5035](https://sorbet.org/docs/error-reference#5035) and [5067](https://sorbet.org/docs/error-reference#5067) as they do not make sense for partial definitions.

### DIFF
--- a/sorbet/config
+++ b/sorbet/config
@@ -1,0 +1,2 @@
+rbi/
+--suppress-error-code=5002,5020,5035,5067
--- a/scripts/check_types
+++ b/scripts/check_types
@@ -15,8 +15,7 @@ files.each do |file|
   $stderr.puts("Typechecking `#{file}`...")
 
   out, status = Open3.capture2e(
-    "bundle exec srb tc #{file} --no-error-count " \
-      "--suppress-error-code=5002,5020,5035,5067"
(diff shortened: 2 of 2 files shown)


KaanOzkan (dismissed):

sorbet/config
```diff
@@ -0,0 +1,2 @@
+rbi/
+--suppress-error-code=5002,5020,5035,5067
```
KaanOzkan, line 2: We can remove this suppress now right https://github.com/Shopify/rbi-central/blob/533bdae78a3c1e6a5b669b9a77e1a3ebca7e4606/scripts/check_types#L19

KaanOzkan (approved):

RyanBrushett (approved):

sorbet/config
Morriar: Actually you made me realize that we should be ignoring the config all together to avoid checking all the files in the same run.

I added `--no-config` to the script call.

Thanks for being so thorough Kaan ❤️

KaanOzkan (approved):

PR #46, opened 2022-06-14
Add RBI for minitest

Morriar:
https://rubygems.org/gems/minitest/versions/5.8.4


Morriar:
I'm going to close this one as most of the signatures contain `T.untyped` and are missing the block definitions.

PR #50, opened 2022-06-22
Check index

Morriar:
A few changes together:

1. Merge `scripts/lint_index_json` into `check_index`
2. Merge `scripts/match_index_entries_with_rbis ` into `check_index`
3. Also validate the index against a json-schema
4. Clean the output

Example of failure: https://github.com/Shopify/rbi-central/runs/7013134492?check_suite_focus=true.


scripts/check_index
```diff
@@ -0,0 +1,104 @@
...
+    @index_json = load_json(index_path)
+    @schema_json = load_json(schema_path)
+  end
+
+  def validate!
+    success = true
+    success = false unless check_against_schema!
+    success = false unless check_missing_index_entries!
+    success = false unless check_missing_annotations!
+    success = false unless check_json_format!
```
st0012, line 34: Isn't this the same as 

```rb
check_against_schema! && check_missing_index_entries! && check_missing_annotations! && check_json_format!
```

I feel assigning booleans based on the same boolean result consumes unnecessary mental computation.

scripts/check_index
```diff
@@ -0,0 +1,104 @@
...
+    File.write(expected_path, expected_json)
+
+    out, status = Open3.capture2e("diff -u #{@index_path} #{expected_path}")
+    unless status.success?
+      $stderr.puts(" * Formatting errors found in `#{@index_path}`:")
+      $stderr.puts("\n#{out}\n")
+      return false
+    end
+
+    FileUtils.rm(expected_path)
```
st0012, line 86: Should this be in `ensure`?

scripts/check_index
```diff
@@ -0,0 +1,104 @@
...
+    success = false unless check_missing_annotations!
+    success = false unless check_json_format!
+    success
+  end
+
+  private
+
+  def check_against_schema!
+    JSON::Validator.validate!(@schema_json, @index_json)
+    return true
```
st0012, line 42: Redundant `return`?
We're returning `true` in `check_json_format!`  so the style also doesn't match.

scripts/check_index
```diff
@@ -0,0 +1,104 @@
...
+  rescue JSON::Schema::ValidationError => e
+    $stderr.puts(" * #{e.message}\n")
+    return false
```
st0012, line 45: Same as the above.

scripts/check_index
st0012: I see. How about just

```rb
success = check_against_schema!
success = check_missing_index_entries!
success = check_missing_annotations!
success = check_json_format!
success
```

schema.json
```diff
@@ -0,0 +1,6 @@
+{
+  "type": "object",
+  "patternProperties": {
+    "^.*$": { "type": "object" }
+  }
+}
```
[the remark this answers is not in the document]
Morriar: At least it catches things like `{ "rails": 42 }` in the index that we do not catch yet.

scripts/check_index
Morriar: TIL.

Note: I still need the `success = true` to not always return false.

st0012 (approved):
❤️

PR #54, opened 2022-06-28
Test RBI definitions at runtime

Morriar:
Following the comments received on #52, I decided to write a script to at least ensure that the properties we define actually exist at runtime.

The objective is twofold:

1. Make it easier to review the PRs
2. Make it easier to understand the annotations and the code they represent

### Index changes

Each annotation from the `rbi/annotations` folder must be defined in the `index.json` file at the root of this repository:

```json
{
    "gemA": { // bem name
        "dependencies": [ // optional: list of gems that need to be installed to test gemA RBI
            "gemB",
            "gemC"
        ],
        "requires": [ // optional: list of files that need to be required to test gemA RBI
            "file1",
            "file2",
        ]
    }
}
```

See the index [validation schema](schema.json) for more details.

### Runtime checks

The approach is somewhat stupid. For each gem we:

1. Create a temporary directory
2. Add a Gemfile pointing to the gem (and possible dependencies, index key `dependencies`)
3. Add a Ruby script that:
   1. Requires the gem (and possible other requirements, index key `requires`)
   2. Tries to `const_get` each constant defined in the RBI file
   3. Tries to call `instance_method` for each method and attribute accessor (or `method` for singleton methods) in the RBI file

It is possible to allow necessary shims for non-existing runtime definitions by using comment tags:

* `@missing_method` to indicate that a method is delegated to another receiver using `method_missing`
* `@shim` to indicate that a definition doesn't actually exist at runtime but is needed to allow type checking

An example of failing CI can be seen here: https://github.com/Shopify/rbi-central/compare/at-test-example?expand=1

See the README for more information.

### DIFF
--- a/scripts/check_runtime
+++ b/scripts/check_runtime
@@ -1,5 +1,6 @@
 #! /usr/bin/env ruby
 
+require "bundler/setup"
 require "rbi"
 
 require_relative "utils/rbi_validator"
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -28,7 +28,7 @@ unless rbis.empty?
     success = false
   end
   $stderr.puts("Checking RBI files against runtime execution...")
-  unless system("bundle exec scripts/check_runtime #{rbis.join(" ")}")
+  unless system("scripts/check_runtime #{rbis.join(" ")}")
     success = false
   end
 end
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,4 +29,4 @@ jobs:
       - name: Check new RBI files belong to public gems
         run: scripts/run_on_changed_rbis scripts/check_gems_are_public
       - name: Check RBI definitions against runtime execution
-        run: scripts/run_on_changed_rbis "bundle exec scripts/check_runtime"
+        run: scripts/run_on_changed_rbis "scripts/check_runtime"


RyanBrushett (commented):
A couple of text comments, some nitpicks. Code looks good to me though.

scripts/check_runtime
```diff
@@ -0,0 +1,294 @@
...
+            const.method("\#{method_name}")
+          else
+            const.instance_method("\#{method_name}")
+          end
+        rescue NameError => e
+          if const && !singleton && __rbi_repo_respond_to_method_missing?(const)
+            return if allow_missing
+
+            $stderr.puts(" * Missing runtime method `\#{recv_name}\#{singleton ? "." : "#"}\#{method_name}` (defined at \#{rbi_loc})")
+            $stderr.puts("   Note: `\#{method_name}` could be delegated to :method_missing but the RBI definition isn't annotated with `@method_missing`.")
```
KaanOzkan, line 264: Why is it *could be delegated* here and not "is being delegated"? Do you have an example in mind where `method_missing` won't be delegating it and it's something else?

scripts/check_runtime
```diff
@@ -0,0 +1,294 @@
...
+        def __rbi_repo_get_method(recv_name, method_name, rbi_loc, singleton:, allow_missing:)
+          const = __rbi_repo_get_const(recv_name, rbi_loc)
+          return unless const
+
+          if singleton
+            const.method("\#{method_name}")
+          else
+            const.instance_method("\#{method_name}")
+          end
+        rescue NameError => e
```
KaanOzkan, line 259: We could use [method_defined?](https://ruby-doc.org/core-3.1.2/Module.html#method_defined-3F-method) on `const` and `const.singleton_class` so that we don't use exceptions as control flow.

scripts/check_runtime
```diff
@@ -0,0 +1,294 @@
...
+        Open3.capture2e("bundle exec ruby #{TEST_NAME}", chdir: @workdir)
+      end
+    end
+
+    def ruby_string
+      <<~RB
+        $success = true
+
+        def __rbi_repo_get_const(const_name, rbi_loc)
+          Kernel.const_get("\#{const_name}")
```
KaanOzkan, line 243: Similarly here we could use [const_defined?](https://ruby-doc.org/core-3.1.2/Module.html#const_defined-3F-method) to not rely on exceptions.

scripts/check_runtime
Morriar: As we can't be sure the method will be delegated properly without actually running the thing:

```ruby
class Foo
  def foo; end
end

class Bar
  def initialize
    @foo = Foo.new
  end

  def method_missing(name)
    @foo.send(name) if rand(10) > 5
  end
end

Bar.new.foo # ok
Bar.new.foo # error
Bar.new.bar # error
```

scripts/check_runtime
Morriar: Since I have to implement the `rescue` for both `__rbi_repo_get_const` and `__rbi_repo_respond_to_method_missing` it made more sense to keep it here as well.

scripts/check_runtime
```diff
@@ -0,0 +1,294 @@
+#! /usr/bin/env ruby
+
```
rafaelfranca, line 2:
```suggestion
require "bundler/setup"
```

And remove `bundle exec` from the execution.

PR #57, opened 2022-06-29
Add static tests against Tapioca and Sorbet

Morriar:
Another PR to make it easier to review PRs:

* Spot redundant definitions with Tapioca generated gems RBIs
* Spot incompatible definitions with Tapioca generated gems RBIs
* Spot type errors more thoroughly

For each gem the test works as follows:

1. Create a temporary directory
2. Add a Gemfile pointing to the gem (and possible dependencies, listed in index key `dependencies`)
3. Generate the RBIs for gems using `tapicoa gems`
4. Check duplication with RBIs from gems using `tapicoa check-shims`
5. Check for type errors using `srb tc` (Sorbet can now be ran without ignoring any error code)

Example of failure: https://github.com/Shopify/rbi-central/tree/at-test-static-example

### DIFF
--- a/index.json
+++ b/index.json
@@ -32,6 +32,8 @@
   },
   "rails": {
     "requires": [
+      "rake",
+      "rake/tasklib",
       "rails/all"
     ]
   },


scripts/check_static
```diff
@@ -0,0 +1,98 @@
...
+  class StaticContext < Context
+    def initialize(repo_index, file)
+      super
+
+      @requires = String.new
+    end
+
+    def run!
+      add_gem_dependency("sorbet", version: ">= 0.5.10109")
+      add_gem_dependency("tapioca", github: "Shopify/tapioca", ref: "8ce7951f69aa41ce9ff4990b67f0c3c9d64c0a6f")
```
rafaelfranca, line 41: Is this ref going to be static?

PR #59, opened 2022-06-29
Add `check_all` script

Morriar:
A script to run all the tests at once.

Usage:

```
# Check all tests on all files
./scripts/check_all 

# Check all tests on specified files
./scripts/check_all rbi/annotations/a.rbi rbi/annotations/b.rbi
```

All tests will be executed regardless of previous failures. That means that you still have the result of `check_runtime` even if `check_index` failed for example.

Also updated the git hook and CI to use it.

Example of failure: https://github.com/Shopify/rbi-central/compare/at-check-all-example?expand=1

### DIFF
--- a/scripts/check_all
+++ b/scripts/check_all
@@ -1,0 +1,27 @@
+#! /usr/bin/env ruby
+
+require "bundler/setup"
+
+require_relative "utils/rbi_validator"
+
+success = true
+
+$stderr.puts("### Checking index...\n\n")
+success &= system("scripts/check_index")
+
+files = rbi_files.join(" ")
+unless files.empty?
+  $stderr.puts("\n### Linting RBI files...\n\n")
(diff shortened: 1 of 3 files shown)


scripts/check_all
```diff
@@ -0,0 +1,27 @@
...
+
+$stderr.puts("### Checking index...\n\n")
+success &= system("scripts/check_index")
+
+files = rbi_files.join(" ")
+unless files.empty?
+  $stderr.puts("\n### Linting RBI files...\n\n")
+  success &= system("bundle exec rubocop #{files}")
+
+  $stderr.puts("\n### Checking all RBI files belong to public gems...\n\n")
```
KaanOzkan, line 17: Nitpick
```suggestion
  $stderr.puts("\n### Checking that all RBI files belong to public gems...\n\n")
```

KaanOzkan (approved):

PR #61, opened 2022-06-29
Add RBI for `configs`

Morriar:
Signed-off-by: Alexandre Terrasa <alexandre.terrasa@shopify.com>

### DIFF
--- a/index.json
+++ b/index.json
@@ -18,6 +18,8 @@
   },
   "colorize": {
   },
+  "configs": {
+  },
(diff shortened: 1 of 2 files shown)


KaanOzkan (commented):
I assume CI is failing because of missing `ActiveSupport`. Can we return the error message before exiting on CI?

Morriar:
It was failing because of #63.

KaanOzkan (approved):

PR #84, opened 2022-07-15
Correct Lhm::Printer::Dot#notify arguments

vinistock:
The [actual implementation](https://github.com/Shopify/lhm/blob/636fb465da29c5f4e6c5d55ba2393f46905018e5/lib/lhm/printer.rb#L54) uses just a `*` argument. The mismatch is causing conflicts with the RBI generated by Tapioca.


st0012 (approved):

vinistock:
Actually, this is not the correct solution. Apologies.

PR #88, opened 2022-07-23
Add annotations for the `aws-sdk-acm` gem

johnernaut:
This gem has a large number of types so this will most likely be incremental, but the types added here so far are
necessary for the project I'm currently working on.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -47,6 +47,7 @@ GEM
 
 PLATFORMS
   arm64-darwin-21
+  ruby
   x86_64-linux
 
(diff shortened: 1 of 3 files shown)


johnernaut:
Thanks for the info @KaanOzkan @Morriar - I attempted to declare `Elem` but to be honest I wasn't really sure how exactly that should be defined since I couldn't find examples of its usage.

johnernaut:
@Morriar Changes made.  Thanks so much!

Morriar (approved):
Thanks!

PR #89, opened 2022-07-25
Add Github PR and Issue templates

wildmaples:
### Motivation

https://github.com/Shopify/rbi-central/issues/37

### Implementation

I used `tapioca` and `sorbet`'s templates as reference. 

### Tests

N/A

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,0 +1,14 @@
+### Problem
+
+<!-- Describe the problem you're encountering, give examples using code if you can -->
+
+### Context
+
+<!-- Provide the following information so we can reproduce the problem -->
+
+* Gem name: <!-- Add the name of the gem you have a problem with here -->
+* Gem version: <!-- Add the version of the gem you have a problem with here (if applicable) -->
+* Gem source: <!-- Link relevant source code from the gem for the problem you have -->
+* Gem API doc: <!-- Link relevant API doc from the gem for the problem you have -->
+* Tapioca version: <!-- Add the version of Tapioca you use -->
+* Sorbet version: <!-- Add the version of Sorbet you use -->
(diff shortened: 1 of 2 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,9 @@
+### Motivation
```
Morriar, line 1: I don't think this template applies for this repo.

What we're interested in when someone pushes a new RBI is the following:

* What is the gem being pushed?
* Where can we find the source code of this gem?
* Where can we find the api documentation of this gem?

If it's a change to an existing gem:

* What was the problem before the change?
* Can the problem be shown with an example or a sorbet.run?
* If you're adding/removing a definition, what is the change in the gem upstream that requires a change here?

etc.

Furthermore, there is not real "implementation", it's just a RBI file and the tests are automated, nothing to push here either.

.github/pull_request_template.md
```diff
@@ -0,0 +1,16 @@
+### Type of Change
+<!-- Select the option that best reflect your changes -->
+<!-- If it's "Other", provide additional detail -->
+- [ ] Add RBI for a new gem
+- [ ] Modify RBI for an existing gem
+- [ ] Other
```
Morriar, line 6:
```suggestion
### Type of Change

<!-- Select the option that best reflect your changes -->

- [ ] Add RBI for a new gem
- [ ] Modify RBI for an existing gem
- [ ] Other: <!-- provide additional information -->
```

.github/issue_template.md
```diff
@@ -0,0 +1,5 @@
+### Problem
+<!-- Describe the problem you're encountering. -->
```
Morriar, line 2:
```suggestion

<!-- Describe the problem you're encountering. -->
```

st0012 (approved):

Morriar (approved):

PR #90, opened 2022-07-29
Extract check scripts as a proper gem

Morriar:
Making this a real gem means we can install it from other repositories (ie. private ones) without copying the files. It makes updates easier.

Next steps would be to make it use Thor for the CLI and maybe some typing/cleaning.


gem/lib/rbi-central/index.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module RBICentral
+  INDEX_PATH = "index.json"
```
st0012, line 4: Do we need to change this to something like `../index.json` or move `index.json` under `gem`?

gem/lib/rbi-central/rbi_validator.rb
```diff
@@ -0,0 +1,38 @@
...
+      @repo_index = repo_index
+      @rbi_file = rbi_file
+      @gem_name = File.basename(rbi_file, ".rbi")
+    end
+
+    def validate_file!; end
+  end
+
+  def self.rbi_files
+    ARGV.empty? ? Dir.glob("./#{RBIS_PATH}/*.rbi") : ARGV
```
st0012, line 36: Does the path need to be adjusted?

gem/lib/rbi-central/cli.rb
```diff
@@ -0,0 +1,91 @@
...
+      $stderr.puts Colors.highlight(message, default_color: Colors::WHITE, highlight_color: Colors::BLUE)
+    end
+
+    def success(message)
+      $stderr.puts message.green
+    end
+  end
+end
+
+class String
```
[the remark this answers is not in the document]
Morriar: No it's only meant to be installed on RBI repositories as a way to check the RBIs they contain.

Also, this will be removed in a subsequent PR where I'm migrating the scripts to be Thor commands.

gem/lib/rbi-central/index.rb
Morriar: No the commands are meant to be run from the repo root directory through `bundle exec <command>` so the index is going to be at the same path.

gem/lib/rbi-central/rbi_validator.rb
Morriar: Same than `index.json`. The command is ran from the repo's root.

st0012 (approved):
👍

PR #91, opened 2022-07-29
Rails.env returns an EnvironmentInquirer

fsateler:
Otherwise `Rails.env.development?` fails to typecheck

### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem name: railties
* Gem version: 6.0.5
* Gem source: https://github.com/rails/rails
* Gem API doc: N/A
* Tapioca version: 0.9.2
* Sorbet version: 0.5.10201


Morriar (approved):
Thanks!

PR #92, opened 2022-08-02
Add Sorbet to `gem/`

Morriar:
### Type of Change

- [ ] Add RBI for a new gem
- [x] Other: CI scripts

### Changes

Add Sorbet annotations to the scripts and enable Rubocop.


gem/lib/rbi-central/context.rb
```diff
@@ -0,0 +1,77 @@
...
+      File.write("#{@workdir}/Gemfile", <<~GEMFILE)
+        source "https://rubygems.org"
+
+        #{@gemfile}
+      GEMFILE
+    end
+
+    sig { returns([String, Process::Status]) }
+    def bundle_install!
+      exec!("bundle config set --local path 'vendor/bundle'")
```
adisonlampert, line 66: Why install to `vendor/bundle`?

gem/sorbet/tapioca/config.yml
```diff
@@ -0,0 +1,13 @@
+gem:
```
adisonlampert, line 1: What is the point of this file?

adisonlampert (approved):

vinistock:
I know this isn't the point of the PR, but I can't help but wonder if adding all of this structure is too much for some RBI checking scripts. Would it make sense for these scripts to belong in `rbi` and would that help us avoid having a new gem?

gem/exe/check_index
```diff
@@ -0,0 +1,118 @@
+#! /usr/bin/env ruby
+# typed: strict
+# frozen_string_literal: true
+
+require "rbi-central"
+
+module RBICentral
+  SCHEMA_PATH = "schema.json"
```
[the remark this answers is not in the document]
Morriar: The schema is used to check the validity of the index object content. It's based on [`json-schema`](https://json-schema.org/).

gem/lib/rbi-central/context.rb
Morriar: To avoid depending on the user configuration. For example to avoid installing things in the system if the user is `root`.

PR #98, opened 2022-08-17
Check runtime ancestors used in annotations files

Morriar:
### Type of Change

- [ ] Add RBI for a new gem
- [x] Other: Ci

### Changes

When running runtime checks we also check that super classes and modules included/extended in the RBI file reflect the runtime behavior.

This allowed me to find a few errors in the current set of annotations (see discussions below).

Closes https://github.com/Shopify/rbi-central/issues/81.

### DIFF
--- a/index.json
+++ b/index.json
@@ -109,7 +109,9 @@
       "minitest"
     ],
     "requires": [
-      "minitest/test"
+      "minitest/test",
+      "webmock/api",
+      "webmock/minitest"
     ]
   }
 }
--- a/rbi/annotations/actionview.rbi
+++ b/rbi/annotations/actionview.rbi
@@ -3,7 +3,7 @@
 module ActionView
   TemplateError = T.type_alias { Template::Error }
 
-  class MissingTemplate < ActionViewError
+  class MissingTemplate < ActionView::ActionViewError
     sig { returns(String) }
     def path; end
   end
--- a/rbi/annotations/lhm-shopify.rbi
+++ b/rbi/annotations/lhm-shopify.rbi
@@ -69,7 +69,7 @@ class Lhm::Printer::Base
   def initialize; end
 end
 
-class Lhm::Printer::Percentage < Lhm::Printer::Base
+class Lhm::Printer::Percentage
   sig { void }
   def end; end
 
--- a/rbi/annotations/parse-cron.rbi
+++ b/rbi/annotations/parse-cron.rbi
@@ -2,8 +2,6 @@
 
 # Parses cron expressions and computes the next occurence of the "job"
 class CronParser
-  extend T::Generic
-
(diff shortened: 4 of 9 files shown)


rbi/annotations/lhm-shopify.rbi
```diff
@@ -69,7 +69,7 @@ class Lhm::Printer::Base
   def initialize; end
 end
 
-class Lhm::Printer::Percentage < Lhm::Printer::Base
+class Lhm::Printer::Percentage
```
Morriar, line 72: See https://github.com/Shopify/lhm/blob/master/lib/lhm/printer.rb#L15

rbi/annotations/parse-cron.rbi
```diff
@@ -71,8 +69,6 @@ end
 
 # internal "mutable" time representation
 class CronParser::InternalTime
-  extend T::Generic
```
Morriar, line 74: See https://github.com/siebertm/parse-cron/blob/master/lib/cron_parser.rb (not generic)

rbi/annotations/parse-cron.rbi
```diff
@@ -2,8 +2,6 @@
 
 # Parses cron expressions and computes the next occurence of the "job"
 class CronParser
-  extend T::Generic
```
Morriar, line 5: See https://github.com/siebertm/parse-cron/blob/master/lib/cron_parser.rb (not generic)

rbi/annotations/actionview.rbi
```diff
@@ -3,7 +3,7 @@
 module ActionView
   TemplateError = T.type_alias { Template::Error }
 
-  class MissingTemplate < ActionViewError
+  class MissingTemplate < ActionView::ActionViewError
```
Morriar, line 6: The problem here is that the class is not flattened properly and thus the superclass name is not fully qualified. We could be smarter but I think it's a good thing to enforce fully qualified names

rbi/annotations/rainbow.rbi
```diff
@@ -15,7 +15,7 @@ module Rainbow
     sig { params(hex: String).returns([Integer, Integer, Integer]) }
     def self.parse_hex_color(hex); end
 
-    class Indexed < Color
+    class Indexed < Rainbow::Color
```
Morriar, line 18: Same here, the class is not flattened properly so the names are not fully qualified.

index.json
```diff
@@ -109,7 +109,9 @@
       "minitest"
     ],
     "requires": [
-      "minitest/test"
+      "minitest/test",
```
Morriar, line 112: Those requires where required to be able to load the constants used in the annotations file.

KaanOzkan (approved):
<3

vinistock (approved):

PR #101, opened 2022-08-29
`Rails.application.config` is untyped

rafaelfranca:
This method returns a hash like object that responds to any method,
if there is a key with that method name.

Since we can't know statically what methods it responds to, we should
not be returning a `Configuration` object.

### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem name: rails
* Gem version: all
* Gem source: https://github.com/rails/rails/blob/39413de44c0e2c0dd2d964be5985b03d8f968a7b/railties/lib/rails/railtie/configuration.rb#L90

### DIFF
--- a/rbi/annotations/railties.rbi
+++ b/rbi/annotations/railties.rbi
@@ -30,21 +30,9 @@ class Rails::Application < ::Rails::Engine
   def config; end
 end
 
-module Rails::Command::Behavior
-  mixes_in_class_methods ::Rails::Command::Behavior::ClassMethods
-end
-
 class Rails::Engine < ::Rails::Railtie
   sig { returns(ActionDispatch::Routing::RouteSet) }
   def routes(&block); end
-end
-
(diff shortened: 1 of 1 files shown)


KaanOzkan (approved):

rbi/annotations/railties.rbi
```diff
@@ -26,27 +26,15 @@ module Rails
 end
 
 class Rails::Application < ::Rails::Engine
-  sig { returns(Rails::Application::Configuration) }
+  sig { returns(T.untyped) }
   def config; end
 end
 
-module Rails::Command::Behavior
```
paracycle, line 33: Are these being removed because Tapioca can generate them?

rafaelfranca: Yes. At least this is what the CI on the previous commit told me.

PR #102, opened 2022-08-29
Include SchemaStatements under ActiveRecord::Migration

KaanOzkan:
Exposes migrations methods that are dynamically mixed in. This is an existing workaround for apps that have migrations marked `typed: true`: https://github.com/Shopify/tapioca/issues/696#issuecomment-1145271363

### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem name:  rails
* Gem source:  https://github.com/rails/rails/blob/f95c0b7e96eb36bc3efc0c5beffbb9e84ea664e4/activerecord/lib/active_record/connection_adapters/abstract/schema_statements.rb#L606

### DIFF
--- a/rbi/annotations/activerecord.rbi
+++ b/rbi/annotations/activerecord.rbi
@@ -4,3 +4,8 @@ class ActiveRecord::Schema
   sig {params(info: T::Hash[T.untyped, T.untyped], blk: T.proc.bind(ActiveRecord::Schema).void).void}
   def self.define(info = nil, &blk); end
 end
+
+class ActiveRecord::Migration
+  # @shim: Methods on migration are delegated to `SchemaStatements` using `method_missing`
+  include ActiveRecord::ConnectionAdapters::SchemaStatements
+end


rbi/annotations/activerecord.rbi
```diff
@@ -4,3 +4,7 @@ class ActiveRecord::Schema
   sig {params(info: T::Hash[T.untyped, T.untyped], blk: T.proc.bind(ActiveRecord::Schema).void).void}
   def self.define(info = nil, &blk); end
 end
+
+class ActiveRecord::Migration
+  include ActiveRecord::ConnectionAdapters::SchemaStatements
```
Morriar, line 9: It doesn't seem this include actually happens at runtime: [`https://github.com/Shopify/rbi-central/runs/8079256178?check_suite_focus=true#step:4:](https://github.com/Shopify/rbi-central/runs/8079256178?check_suite_focus=true#step:4:40) where is it done in the code?

KaanOzkan: I think this is where migration methods are being sent to the appropriate DB adapter https://github.com/rails/rails/blob/39413de44c0e2c0dd2d964be5985b03d8f968a7b/activerecord/lib/active_record/migration.rb#L946.

So a more accurate solution would be to list the methods defined under `SchemaStatements`. However, there's quite a few of them 😕. We could just do popular ones like `add_column`, `add_index`.

Morriar: If we really want to go this route we should implement [@shim](https://github.com/Shopify/rbi-central#:~:text=%40shim%20to%20indicate%20that%20a%20definition%20doesn%27t%20actually%20exist%20at%20runtime%20but%20is%20needed%20to%20allow%20type%20checking) on `include` so we have a way to relax the checks for a particular include and we can attach documentation to it explaining why we made this choice.

KaanOzkan: Does it need an implementation? Would `include` calls not hit https://github.com/Shopify/rbi-central/blob/91ec8bdaa0e65824184230600534a58492e974a1/gem/lib/rbi-central/runtime/visitor.rb#L32?

Morriar (approved):

PR #103, opened 2022-08-31
Adds shim for `sorbet-coerce`

mattxwang:
### Type of Change

- [x] Add RBI for a new gem
- [ ] Other: 

### Changes

I'm an interim maintainer for  [`sorbet-coerce`](https://github.com/chanzuckerberg/sorbet-coerce) and am adding a shim! This is what we use internally @chanzuckerberg. This shim is valid for the latest version of `sorbet-coerce` and works across various Tapioca/Sorbet versions.

I did want to ask about including `Elem = type member`. In our internal codebase, including that line causes a duplicate definition of the `type_member`; however, if I remove that line, then `bundle exec repo check static sorbet-coerce` fails with 

```console
$ bundle exec repo check static sorbet-coerce 
Error: rbi/annotations/sorbet-coerce.rbi:5: Unable to resolve constant Elem https://srb.help/5002
     5 |  sig { params(args: T.untyped, raise_coercion_error: T.nilable(T::Boolean), coerce_empty_to_nil: T::Boolean).returns(Elem) }
```

which locally no longer exists in the file, but should be defined by Tapioca's generated file from `bin/tapioca gems`. How should I proceed here?

Note: I used a slightly different definition for sorbet-coerce on sorbet-typed: https://github.com/sorbet/sorbet-typed/pull/396.

### DIFF
--- a/index.json
+++ b/index.json
@@ -100,6 +100,8 @@
       "sidekiq/web"
     ]
   },
+  "sorbet-coerce": {
+  },
   "state_machines": {
   },
   "stripe": {
--- a/rbi/annotations/sorbet-coerce.rbi
+++ b/rbi/annotations/sorbet-coerce.rbi
@@ -1,0 +1,9 @@
+# typed: strict
+
+module TypeCoerce
+  Elem = type_member
+
+  # @shim: at runtime, this is delegated to TypeCoerce::Converter; the use-case looks like: TypeCoerce[Klass].new.from(...)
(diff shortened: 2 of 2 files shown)


mattxwang:
Hm ... I'm now second-guessing this type signature (since it doesn't play as nicely with tapioca's sorbet-coerce.rbi). Perhaps don't merge yet?

mattxwang:
Thanks for the suggestion @Morriar! I *think* we did have RBIs directly within the gem, but I'm not sure if they work properly with tapioca. Let me look into that and see if it's possible. Appreciate the suggestion, I'm definitely not very familiar with best practices on maintaining sorbet-related gems!

PR #107, opened 2022-09-18
Improve ActiveRecord::Migration types by including DatabaseStatements module.

connorshea:
### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem name: activerecord
* Gem version: 7.0.4
* Gem source: [This inclusion was not represented](https://github.com/rails/rails/blob/fbc4dff57866ac35e000b86a51afe28ec1606a8f/activerecord/lib/active_record/connection_adapters/abstract_adapter.rb#L33)
* Gem API doc: [the `execute` method was not being understood in migrations](https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/DatabaseStatements.html#method-i-execute)
* Tapioca version: 0.10.1
* Sorbet version: 0.5.10439

There are various methods that were missing due to the lack of the DatabaseStatements module.

You can see the inclusion here in the Rails repo:
https://github.com/rails/rails/blob/fbc4dff57866ac35e000b86a51afe28ec1606a8f/activerecord/lib/active_record/connection_adapters/abstract_adapter.rb#L33

### DIFF
--- a/rbi/annotations/activerecord.rbi
+++ b/rbi/annotations/activerecord.rbi
@@ -8,5 +8,6 @@ end
 class ActiveRecord::Migration
   # @shim: Methods on migration are delegated to `SchemaStatements` using `method_missing`
   include ActiveRecord::ConnectionAdapters::SchemaStatements
+  # @shim: Methods on migration are delegated to `DatabaseaStatements` using `method_missing`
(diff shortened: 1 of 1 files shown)


connorshea:
Updated with the `@shim` 👍

PR #110, opened 2022-09-20
add annotation for sidekiq-scheduler

ghost:
### Type of Change

- [x] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem name: `sidekiq-scheduler`
* Gem version: `4.0.2`
* Gem source: https://github.com/sidekiq-scheduler/sidekiq-scheduler/blob/8ad1f746b2743c13ba9b37309271bd41aff2a4f1/lib/sidekiq-scheduler/scheduler.rb#L39-L41
* Gem API doc: 
* Tapioca version: `0.10.1`
* Sorbet version: `0.5.10426`

This solution came from a Discourse post here https://discourse.shopify.io/t/method-reload-schedule-does-not-exist-on-t-class-of-sidekiqscheduler-scheduler/17836 

The rbi shim changed slightly since that post though. Ran into this issue again converting a new repo to `typed: true`

### DIFF
--- a/rbi/annotations/sidekiq-scheduler.rb
+++ b/rbi/annotations/sidekiq-scheduler.rb
@@ -1,7 +1,9 @@
-# typed: true
+# typed: strong
 
 class SidekiqScheduler::Scheduler
+  sig { void }
   def self.reload_schedule!; end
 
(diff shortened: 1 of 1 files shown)


oneandonlymike:
> This looks good to me, but can we please add signatures to the methods?

Added 😄

vinistock (approved):

PR #111, opened 2022-09-21
add sidekiq-scheduler to index

ghost:
Just a quick follow up to https://github.com/Shopify/rbi-central/pull/110 where I forgot to add my new gem annotation to the index so it wasn't being fetched


vinistock (approved):

oneandonlymike:
And if I had added it to the index the first time, it would have caught that I had accidentally used `rb` instead of `rbi` on the annotation 🤦

vinistock:
It seems our checks are failing because we're not finding the method in the runtime. Do you need to do anything for these methods to exist?

PR #114, opened 2022-09-30
The block passed to field gets a Field param

kbarrette:
### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem name: graphql
* Gem version: 
* Gem source: 
* Gem API doc: 
* Tapioca version: 
* Sorbet version:

### DIFF
--- a/rbi/annotations/graphql.rbi
+++ b/rbi/annotations/graphql.rbi
@@ -49,7 +49,7 @@ module GraphQL::Schema::Member::HasFields
     params(
       args: T.untyped,
       kwargs: T.untyped,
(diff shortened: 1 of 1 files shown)


KaanOzkan:
Does it have to be? Looks like that block is passed to `initialize` as `definition_block` which can take multiple number of arguments. https://github.com/rmosolgo/graphql-ruby/blob/2515f2ad938265c7e3adc6d9f127539fd15bd76a/lib/graphql/schema/field.rb#L340-L345. If that's the case we should make it an optional argument.

PR #119, opened 2022-10-20
Fix aws-sdk-acm runtime check

thomasmarshall:
### Type of Change

- [ ] Add RBI for a new gem
- [x] Other: The aws-sdk-acm gem requires a compatible XML library to be installed.


Morriar (approved):
Thanks!

PR #120, opened 2022-10-20
Allow @method_missing annotation for singleton methods

### LINKED ISSUE #118: Validate `@missing_method` for singleton methods.
### Problem

Currently, we skip the validation of `@missing_method` if the method is a singleton method: https://github.com/Shopify/rbi-central/blob/main/gem/lib/rbi-central/runtime/context.rb#L206 but this case can actually happen: https://github.com/sidekiq-scheduler/sidekiq-scheduler/blob/8ad1f746b2743c13ba9b37309271bd41aff2a4f1/lib/sidekiq-scheduler/scheduler.rb#L39-L41. We should validate in all cases.

thomasmarshall:
### Type of Change

- [ ] Add RBI for a new gem
- [x] Other: Fixes #118.

This PR ensures the `@method_missing` annotation works on singleton methods by validating that the `method_missing` method is defined.

### DIFF
--- a/gem/lib/rbi-central/runtime/context.rb
+++ b/gem/lib/rbi-central/runtime/context.rb
@@ -203,7 +203,7 @@ module RBICentral
               const.instance_method("\#{method_name}")
             end
           rescue NameError => e
-            if const && !singleton && __rbi_repo_respond_to_method_missing?(const)
+            if const && __rbi_repo_respond_to_method_missing?(const, singleton: singleton)
               return if allow_missing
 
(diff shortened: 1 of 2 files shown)


thomasmarshall:
@Morriar I've rebased now, so those CI errors are passing.

There was another issue with how `method(:method_missing)` behaved in Ruby 2.7 (in CI) compared to Ruby 3.1 (on my development machine), which I addressed by using `singleton_method` instead. I think that's fine because we know we're looking for a singleton method, but maybe there's some subtle behavior that's not yet apparent to me.

PR #130, opened 2022-11-18
Update RBI for mocha 2.0

olivier-thatch:
### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem name: mocha
* Gem version: 2.0.2
* Gem source: https://github.com/freerange/mocha
* Gem API doc: https://github.com/freerange/mocha#usage
* Tapioca version: Tapioca v0.10.3
* Sorbet version: Sorbet typechecker 0.5.10559 git b5a80d8c764a77dba6409f40116121cd13294a10 debug_symbols=true clean=1

Additional notes:
- this fixes the signature to work with mocha 2.0.2, but it will likely break it for people still running mocha 1.x. I assume that the annotations in this repo are meant to work with the most recent version of their gems, but let me know if that's incorrect
- The `kwargs` argument is added because mocha calls `ruby2_keywords` on the method [here](https://github.com/freerange/mocha/blob/v2.0.2/lib/mocha/expectation.rb#L273), cf. #128 and https://github.com/Shopify/tapioca/issues/1273#issuecomment-1320490109


olivier-thatch:
I've just submitted the CLA form as well.

dirceu:
@KaanOzkan I do get the same error from CI when I run tests locally 🤔

olivier-thatch:
Ah yeah, that is unfortunate. 👍 to removing the annotation, I don't really see another way around until Sorbet/Tapioca natively support `ruby2_keywords` and can perform the necessary signature tweaking internally.

dirceu:
@olivier-thatch can you please rebase with `main` so we can get a clean CI? We've updated CI to focus on Ruby 3.1 and unblock this while we look for a better way to preserve compatibility with multiple versions.

dirceu (approved):

PR #134, opened 2022-12-09
Add RBI for GlobalID

bdewater:
### Type of Change

- [x] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem name: GlobalID
* Gem source: https://github.com/rails/globalid


index.json
```diff
@@ -50,6 +50,14 @@
   },
   "faraday": {
   },
+  "globalid": {
+    "dependencies": [
+      "activerecord"
```
KaanOzkan, line 55: We don't actually need this because missing constant errors are turned off on CI.

KaanOzkan (approved):

index.json
bdewater: TIL - removed it! thanks :)

rbi/annotations/globalid.rbi
```diff
@@ -0,0 +1,6 @@
+# typed: strict
+
+class ActiveRecord::Base
+  # @shim: this is included at runtime https://github.com/rails/globalid/blob/5da635575f0f1f0251778dc21bf3b960e63b5791/lib/global_id/railtie.rb#L38
```
Morriar, line 4: I think we can point to the source of 1.0.0 rather than a specific commit?

https://github.com/rails/globalid/blob/v1.0.0/lib/global_id/railtie.rb#L38

bdewater-thatch: Sure!

KaanOzkan:
Thanks Bart. I'm merging this with a CI failure because it's unrelated to the annotations and requires a new tapioca version. 

An alternative is fixing the Sorbet version but I like that we use the latest one. It allows us to know about any compatibility issues early.

PR #136, opened 2022-12-12
Do not sign commits during tests

KaanOzkan:
### Type of Change

- [ ] Add RBI for a new gem
- [x] Other: 

### Changes

GPG signing clutters the test output and is probably slower. It's done by default in Shopify machines so I think it's worth disabling it in our tests.

Using https://git-scm.com/docs/git#Documentation/git.txt--cltnamegtltvaluegt


Morriar (approved):
Good idea!

I think you can also apply this to Spoom: https://github.com/Shopify/spoom/blob/main/test/test_project.rb#L16-L20

PR #140, opened 2022-12-16
Add RBI for Kredis

bdewater-thatch:
### Type of Change

- [x] Add RBI for a new gem
- [ ] Other: 

`Kredis::Attributes` is also [included in `ActiveModel::Model`](https://github.com/rails/kredis/blob/v1.3.0/lib/kredis/railtie.rb#L25-L27), but I had trouble getting the requires right (it seemed like a game of whack-a-mole adding requires for model, api, validations, etc in index.json), any pointers welcome on how to do this in a more sane way.

### Changes

* Gem name: Kredis
* Gem version: 1.3.0
* Gem source: https://github.com/rails/kredis/blob/v1.3.0/lib/kredis/types.rb
* Tapioca version: 0.10.3
* Sorbet version: 0.5.10565 git cd9e60b894bd7de78c100ea91ad8dd20f4109ee2 debug_symbols=true clean=1

### DIFF
--- a/index.json
+++ b/index.json
@@ -60,6 +60,16 @@
   },
   "graphql": {
   },
+  "kredis": {
+    "dependencies": [
+      "activemodel",
+      "kredis"
+    ],
+    "requires": [
+      "active_model",
+      "kredis"
+    ]
+  },
   "lhm": {
   },
   "lhm-shopify": {
--- a/rbi/annotations/kredis.rbi
+++ b/rbi/annotations/kredis.rbi
@@ -1,0 +1,133 @@
+# typed: strict
+
+module Kredis::Types
+  sig { params(key: T.untyped, default: T.untyped, config: T.untyped, after_change: T.untyped, expires_in: T.untyped).returns(Kredis::Types::Scalar) }
+  def boolean(key, default: nil, config: nil, after_change: nil, expires_in: nil); end
+
+  sig { params(key: T.untyped, expires_in: T.untyped, config: T.untyped, after_change: T.untyped).returns(Kredis::Types::Counter) }
+  def counter(key, expires_in: nil, config: nil, after_change: nil); end
+
+  sig { params(key: T.untyped, values: T.untyped, expires_in: T.untyped, config: T.untyped, after_change: T.untyped).returns(Kredis::Types::Cycle) }
+  def cycle(key, values:, expires_in: nil, config: nil, after_change: nil); end
+
+  sig { params(key: T.untyped, default: T.untyped, config: T.untyped, after_change: T.untyped, expires_in: T.untyped).returns(Kredis::Types::Scalar) }
+  def datetime(key, default: nil, config: nil, after_change: nil, expires_in: nil); end
+
+  sig { params(key: T.untyped, default: T.untyped, config: T.untyped, after_change: T.untyped, expires_in: T.untyped).returns(Kredis::Types::Scalar) }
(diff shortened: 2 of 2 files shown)


index.json
```diff
@@ -53,13 +53,22 @@
   "globalid": {
     "dependencies": [
       "activerecord"
-     ],
+    ],
```
bdewater-thatch, line 56: This snuck in from https://github.com/Shopify/rbi-central/pull/134 and it seems CI only checks the index when the file has changed and not all the time, therefore main is green ([see eg this merge](https://github.com/Shopify/rbi-central/actions/runs/3696918484/jobs/6261262917))

rbi/annotations/kredis.rbi
```diff
@@ -0,0 +1,183 @@
...
+end
+
+class Kredis::Types::Proxy
+  # @method_missing: subclasses delegate (via `proxying` class method) to Proxy, https://github.com/rails/kredis/blob/v1.3.0/lib/kredis/types/proxy.rb#L23
+  sig { params(_arg0: T.untyped, _arg1: T.untyped, _arg2: T.nilable(T.proc.void)).returns(T::Boolean) }
+  def exists?(*_arg0, **_arg1, &_arg2); end
+end
+
+class Kredis::Types::Scalar
+  extend T::Generic
```
bdewater-thatch, line 136: The runtime check doesn't like these declarations, but not sure how to fix. Seems useful to know what values we'll be getting back, this is what happens at runtime:
```ruby
i = Kredis.integer "ex1"
#<Kredis::Types::Scalar:0x0000000109504010                   
 @default=nil,                                               
 @expires_in=nil,                                            
 @key="ex1",                                                 
 @proxy=#<Kredis::Types::Proxy:0x000000010956ff90 @key="ex1", @redis=#<Redis client v5.0.5 for redis://127.0.0.1:6379/0>>,
 @typed=:integer>  
i.value = "42"
# Kredis Proxy (1.4ms)  SET ex1 [42] 
=> "42"                                            
i.value
# Kredis Proxy (1.1ms)  GET ex1  
=> 42
```

KaanOzkan: I don't think we can go this route. Generic is a runtime behaviour that can't be only represented from RBIs.

Having an annotation like this would make the static checker not error for usages such as `Kredis[Integer].new(42)` which is an error because Kredis doesn't actually extend `T::Generic` and won't have access to [#[]](https://github.com/sorbet/sorbet/blob/052762b0658d99d0c622d535b181b93ff7af9267/gems/sorbet-runtime/lib/types/generic.rb#L11).

I think only alternative without changing the library is keeping the signatures as `T.untyped`. Which may mean these annotations aren't necessary. 

cc @Morriar if you have anything to add. I was slightly confused because of https://github.com/Shopify/rbi-central/pull/64

bdewater-thatch: Bummer. In that case I can cut anything using generics, some signatures are still useful here IMO (eg the factory methods for `Kredis::Types`). Will update.

bdewater-thatch:
Updated to remove generics.

I also removed the signature for `Kredis.enum("key")` as this dynamically generates methods and its cumbersome to deal with the setter and predicate methods this RBI cannot know about. The Tapioca compiler will generate these methods when using the `kredis_enum :key` class method for Active Model/Record classes.

KaanOzkan (approved):
LGTM

PR #143, opened 2023-01-09
`ActiveSupport::TimeWithZone` inherits `Time`

tonybruess:
According to documentation, `ActiveSupport::TimeWithZone` implements all `Time` methods. Is this the best way to do this?

https://api.rubyonrails.org/classes/ActiveSupport/TimeWithZone.html

### DIFF
--- a/rbi/annotations/activesupport.rbi
+++ b/rbi/annotations/activesupport.rbi
@@ -46,3 +46,6 @@ class Object
   sig { returns(T::Boolean) }
   def present?; end
 end
+
+class ActiveSupport::TimeWithZone < Time
+end
--- a/rbi/annotations/activesupport.rbi
+++ b/rbi/annotations/activesupport.rbi
@@ -47,5 +47,7 @@ class Object
   def present?; end
 end
 
-class ActiveSupport::TimeWithZone < Time
+class ActiveSupport::TimeWithZone
+   # @shim: Methods are delegated to `Time` using `method_missing`
+  include Time
 end


rbi/annotations/activesupport.rbi
```diff
@@ -46,3 +46,6 @@ class Object
   sig { returns(T::Boolean) }
   def present?; end
 end
+
+class ActiveSupport::TimeWithZone < Time
+end
```
KaanOzkan, line 51: I think representing it as a mixin is slightly more accurate and you can satisfy our CI using `@shim`, similar to https://github.com/Shopify/rbi-central/blob/cd36dd5bc194097e703c384d8e7b94370be45df8/rbi/annotations/activerecord.rbi#L9

KaanOzkan (commented):
Missing methods are being delegated to `Time` so this looks good to me 

https://github.com/rails/rails/blob/8015c2c2cf5c8718449677570f372ceb01318a32/activesupport/lib/active_support/time_with_zone.rb#L567-L573

rbi/annotations/activesupport.rbi
tonybruess: Perf, will change

tonybruess:
Change to a mixin, fixed the build ◡̈

KaanOzkan:
@tonybruess let's identify the methods missing between `Time` and `TimeWithZone` and manually define them please. I was under incorrect assumption that inheritance would be viable but we don't want exactly what it entails. I've added an explanation on the tapioca PR too https://github.com/Shopify/tapioca/pull/1352#issuecomment-1379548571. Sorry about the incorrect back and forth ❤️

PR #144, opened 2023-01-10
Add signatures for ActiveSupport extensions to Array

dirceu:
### Type of Change

### Changes

Add signatures to all extensions for `Array` made by `ActiveSupport`. Closes https://github.com/Shopify/rbi-central/issues/139

* Gem name: activesupport
* Gem version: 7.1.0.alpha
* Gem source: https://github.com/rails/rails/tree/main/activesupport/lib/active_support/core_ext/array

### DIFF
--- a/rbi/annotations/activesupport.rbi
+++ b/rbi/annotations/activesupport.rbi
@@ -46,3 +46,83 @@ class Object
   sig { returns(T::Boolean) }
   def present?; end
 end
+
+class Hash
+  sig { returns(T::Boolean) }
+  def extractable_options?; end
+end
+
+class Array
+  sig { params(position: Integer).returns(T.self_type) }
+  def from(position); end
+
+  sig { params(position: Integer).returns(T.self_type) }
+  def to(position); end
+
+  sig { params(elements: T.untyped).returns(T::Array[T.untyped]) }
+  def including(*elements); end
+
+  sig { params(elements: T.untyped).returns(T.self_type) }
+  def excluding(*elements); end
+
+  sig { params(elements: T.untyped).returns(T.self_type) }
+  def without(*elements); end
+
+  sig { returns(T.nilable(Elem)) }
+  def second; end
+
+  sig { returns(T.nilable(Elem)) }
+  def third; end
+
+  sig { returns(T.nilable(Elem)) }
+  def fourth; end
+
+  sig { returns(T.nilable(Elem)) }
+  def fifth; end
+
+  sig { returns(T.nilable(Elem)) }
+  def forty_two; end
+
+  sig { returns(T.nilable(Elem)) }
+  def third_to_last; end
+
+  sig { returns(T.nilable(Elem)) }
+  def second_to_last; end
+
+  sig { params(options: T::Hash[T.untyped, T.untyped]).returns(String) }
+  def to_sentence(options = {}); end
+
+  sig { params(format: Symbol).returns(String) }
+  def to_fs(format = :default); end
+
+  sig { params(format: Symbol).returns(String) }
+  def to_formatted_s(format = :default); end
+
+  sig { returns(String) }
+  def to_xml; end
+
+  sig { returns(T::Hash[T.untyped, T.untyped]) }
+  def extract_options!; end
+
+  sig do
+    params(number: Integer, fill_with: T.untyped, block: T.nilable(T.proc.void)).returns(T::Array[T::Array[Elem]])
+  end
+  def in_groups(number, fill_with = nil, &block); end
+
+  sig do
+    params(number: Integer, fill_with: T.untyped, block: T.nilable(T.proc.void)).returns(T::Array[T::Array[Elem]])
+  end
+  def in_groups_of(number, fill_with = nil, &block); end
+
+  sig { params(value: T.untyped, block: T.nilable(T.proc.void)).returns(T::Array[T::Array[Elem]]) }
+  def split(value = nil, &block); end
+
+  sig { params(object: T.untyped).returns(T::Array[T.untyped]) }
+  def self.wrap(object); end
+
+  sig { returns(T.untyped) }
+  def extract!; end
+
+  sig { returns(ActiveSupport::ArrayInquirer) }
+  def inquiry; end
+end
--- a/rbi/annotations/activesupport.rbi
+++ b/rbi/annotations/activesupport.rbi
@@ -105,16 +105,31 @@ class Array
   def extract_options!; end
 
   sig do
-    params(number: Integer, fill_with: T.untyped, block: T.nilable(T.proc.void)).returns(T::Array[T::Array[Elem]])
+    type_parameters(:FillType)
+      .params(
+        number: Integer,
+        fill_with: T.type_parameter(:FillType),
+        block: T.nilable(T.proc.params(group: T::Array[T.any(Elem, T.type_parameter(:FillType))]).void),
+      )
+      .returns(T::Array[T.any(Elem, T.type_parameter(:FillType))])
   end
-  def in_groups(number, fill_with = nil, &block); end
+  def in_groups(number, fill_with = T.unsafe(nil), &block); end
 
   sig do
-    params(number: Integer, fill_with: T.untyped, block: T.nilable(T.proc.void)).returns(T::Array[T::Array[Elem]])
+    type_parameters(:FillType)
+      .params(
+        number: Integer,
+        fill_with: T.type_parameter(:FillType),
+        block: T.nilable(T.proc.params(group: T::Array[T.any(Elem, T.type_parameter(:FillType))]).void),
+      )
+      .returns(T::Array[T.any(Elem, T.type_parameter(:FillType))])
   end
-  def in_groups_of(number, fill_with = nil, &block); end
+  def in_groups_of(number, fill_with = T.unsafe(nil), &block); end
 
-  sig { params(value: T.untyped, block: T.nilable(T.proc.void)).returns(T::Array[T::Array[Elem]]) }
+  sig do
+    params(value: T.untyped, block: T.nilable(T.proc.params(element: Elem).returns(T.untyped)))
+      .returns(T::Array[T::Array[Elem]])
+  end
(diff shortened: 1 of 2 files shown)


rbi/annotations/activesupport.rbi
```diff
@@ -46,3 +46,71 @@ class Object
   sig { returns(T::Boolean) }
   def present?; end
 end
+
+class Hash
+  sig { returns(T::Boolean) }
+  def extractable_options?; end
+end
```
paracycle, line 53: Any reason why this is included?

rbi/annotations/activesupport.rbi
```diff
@@ -46,3 +46,71 @@ class Object
...
+
+class Array
+  sig { params(position: Integer).returns(T::Array[T.untyped]) }
+  def from(position); end
```
paracycle, line 57: We should be able to use either `returns(T.self_type)` or `returns(T::Array[Elem])` for this and most of the methods below.

For methods that return the same type as the receiver, `T.self_type` is more appropriate. For methods that return the type of the element, like `second` for example, using `Elem` is more appropriate.

rbi/annotations/activesupport.rbi
```diff
@@ -46,3 +46,71 @@ class Object
...
+
+  sig do
+    params(
+      number: Integer, fill_with: T.untyped, block: T.nilable(T.proc.void)
+    ).returns(T::Array[T.untyped])
+  end
+  def in_groups_of(number, fill_with = nil, &block); end
+
+  sig { returns(T.untyped) }
+  def self.wrap; end
```
paracycle, line 109: `wrap` should take an argument to wrap, no?

rbi/annotations/activesupport.rbi
```diff
@@ -46,3 +46,71 @@ class Object
...
+
+  sig { returns(T.untyped) }
+  def extract!; end
+
+  sig { returns(T.untyped) }
```
paracycle, line 114: This returns an `ArrayInquirer` type
```suggestion
  sig { returns(ActiveSupport::ArrayInquirer) }
```

paracycle (changes requested):
I think we have have this better typed.

Additionally, have we checked this against Core before committing here to make sure that we don't cause a lot of errors when this file is pulled?

dirceu:
@paracycle I've tested this against Shopify Core and changes on this PR look good after some auto-correctable fixes (e.g. changing a bunch of tests that use `.second` to wrap those calls with a `T.must`).

vinistock (approved):

rbi/annotations/activesupport.rbi
```diff
@@ -46,3 +46,69 @@ class Object
...
+
+  sig { returns(T::Hash[T.untyped, T.untyped]) }
+  def extract_options!; end
+
+  sig do
+    params(number: Integer, fill_with: T.untyped, block: T.nilable(T.proc.void)).returns(T::Array[T::Array[Elem]])
+  end
+  def in_groups_of(number, fill_with = nil, &block); end
+
+  sig { params(object: T.untyped).returns(T.untyped) }
```
paracycle, line 106: Doesn't this always return an `Array`?
```suggestion
  sig { params(object: T.untyped).returns(T::Array[T.untyped]) }
```

rbi/annotations/activesupport.rbi
```diff
@@ -46,3 +46,69 @@ class Object
...
+  def third_to_last; end
+
+  sig { returns(T.nilable(Elem)) }
+  def second_to_last; end
+
+  sig { params(options: T::Hash[T.untyped, T.untyped]).returns(String) }
+  def to_sentence(options = {}); end
+
+  sig { params(format: Symbol).returns(String) }
+  def to_fs(format = :default); end
```
paracycle, line 93: I think we are missing:
```ruby
alias_method :to_formatted_s, :to_fs
```

rbi/annotations/activesupport.rbi
```diff
@@ -46,3 +46,69 @@ class Object
...
+  def from(position); end
+
+  sig { params(position: Integer).returns(T.self_type) }
+  def to(position); end
+
+  sig { params(elements: T.untyped).returns(T.self_type) }
+  def including(*elements); end
+
+  sig { params(elements: T.untyped).returns(T.self_type) }
+  def excluding(*elements); end
```
paracycle, line 66: I think we are missing:
```ruby
alias :without :excluding
```

rbi/annotations/activesupport.rbi
```diff
@@ -46,3 +46,69 @@ class Object
...
+end
+
+class Array
+  sig { params(position: Integer).returns(T.self_type) }
+  def from(position); end
+
+  sig { params(position: Integer).returns(T.self_type) }
+  def to(position); end
+
+  sig { params(elements: T.untyped).returns(T.self_type) }
```
paracycle, line 62: Ah, this one is not so cut and dry, since it potentially adds the passed elements to the array, but the passed elements might be a different type to `Elem`, so the array element type might change, and `returns(T.self_type)` might not be correct:
```ruby
[1,2,3].including(:a, :b) #=> [1, 2, 3, :a, :b]
```

paracycle (commented):
I am not seeing the following methods in here:
```ruby
def split(value = nil, &block); end
def in_groups(number, fill_with = nil, &block); end
```

rbi/annotations/activesupport.rbi
dirceu: This (and for the `alias_method` suggestion) cause conflicts when typechecking:

```
sorbet/rbi/annotations/activesupport.rbi:95: Redefining the existing method Array#to_formatted_s as a method alias https://development.shopify.io/engineering/keytech/reference/sorbet/error_reference#5037
    95 |  alias_method :to_formatted_s, :to_fs
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    sorbet/rbi/gems/activesupport@7.1.0.alpha-2da863e0c380ef6bb9ddbd223a301554feb6a40e.rbi:14302: Previous definition
       14302 |  def to_formatted_s(format = T.unsafe(nil)); end
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

I’ve tried regenerating the activesupport RBI but it still has a conflicting definition. 

Because of that, I've added "duplicated" method definitions here instead of using `alias*`

paracycle (commented):
I have a couple of more suggestions, but this is looking really good now.

rbi/annotations/activesupport.rbi
```diff
@@ -46,3 +46,83 @@ class Object
...
+  def to_formatted_s(format = :default); end
+
+  sig { returns(String) }
+  def to_xml; end
+
+  sig { returns(T::Hash[T.untyped, T.untyped]) }
+  def extract_options!; end
+
+  sig do
+    params(number: Integer, fill_with: T.untyped, block: T.nilable(T.proc.void)).returns(T::Array[T::Array[Elem]])
```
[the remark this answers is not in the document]
paracycle: Here is it in action: https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20Array%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7B%20%0A%20%20%20%20type_parameters%28%3AT%29%0A%20%20%20%20%20%20.params%28%0A%20%20%20%20%20%20%20%20number%3A%20Integer%2C%0A%20%20%20%20%20%20%20%20fill_with%3A%20T.type_parameter%28%3AT%29%2C%20%0A%20%20%20%20%20%20%20%20blk%3A%20T.proc%0A%20%20%20%20%20%20%20%20%20%20.params%28%0A%20%20%20%20%20%20%20%20%20%20%20%20elem%3A%20T%3A%3AArray%5BT.any%28Elem%2C%20T.type_parameter%28%3AT%29%29%5D%0A%20%20%20%20%20%20%20%20%20%20%29.void%0A%20%20%20%20%20%20%29%0A%20%20%20%20%20%20.returns%28T%3A%3AArray%5BT.any%28T.type_parameter%28%3AT%29%2C%20Elem%29%5D%29%0A%20%20%7D%0A%20%20def%20in_groups%28number%2C%20fill_with%20%3D%20T.unsafe%28nil%29%2C%20%26blk%29%0A%20%20%20%20T.unsafe%28append%29%20%2B%20%5Bfill_with%5D%20%23%20dummy%20impl%0A%20%20end%0Aend%0A%0AT.reveal_type%28%0A%20%20%5B1%2C%202%2C%203%5D.in_groups%283%2C%20%22as%22%29%20%7B%20%7Cgroup%7C%20%20T.reveal_type%28group%29%20%7D%0A%29%0AT.reveal_type%28%0A%20%20%5B%3Afoo%2C%20%3Abar%5D.in_groups%284%2C%20%22not%20an%20int%22%29%20%7B%20%7Cgroup%7C%20%20T.reveal_type%28group%29%20%7D%0A%29%0AT.reveal_type%28%0A%20%20%5B%3Afoo%2C%20%3Abar%5D.in_groups%284%29%20%7B%20%7Cgroup%7C%20%20T.reveal_type%28group%29%20%7D%0A%29

paracycle (approved):

rbi/annotations/activesupport.rbi
```diff
@@ -46,3 +46,98 @@ class Object
...
+  def extract_options!; end
+
+  sig do
+    type_parameters(:FillType)
+      .params(
+        number: Integer,
+        fill_with: T.type_parameter(:FillType),
+        block: T.nilable(T.proc.params(group: T::Array[T.any(Elem, T.type_parameter(:FillType))]).void),
+      )
+      .returns(T::Array[T.any(Elem, T.type_parameter(:FillType))])
```
Slike9, line 114: I suppose the return type should be `T::Array[T::Array[T.any(Elem, T.type_parameter(:FillType))]]` (another T::Array around).

KaanOzkan: https://github.com/Shopify/rbi-central/pull/145

PR #146, opened 2023-01-27
Add types to GlobalID::Identification

bdewater:
### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other: 

### Changes

I keep using `to_gid` as if it returns a string and it doesn't 😅 https://github.com/rails/globalid/blob/main/lib/global_id/identification.rb

### DIFF
--- a/index.json
+++ b/index.json
@@ -56,7 +56,8 @@
       "activerecord"
     ],
     "requires": [
-      "active_record"
+      "active_record",
+      "global_id/identification"
     ]
   },
   "graphql": {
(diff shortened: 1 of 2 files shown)


Morriar:
The runtime checks do not seem to find `GlobalID::Identification`, do we need to manually require `/global_id/identification` here: https://github.com/Shopify/rbi-central/blob/main/index.json#L59?

bdewater:
Yep, I had that locally but forgot to commit it 🤦‍♂️  CI passes now.

Morriar (approved):

Morriar:
Thanks!

PR #147, opened 2023-02-20
Add VS Code launch config for debug gem

andyw8:
Closes https://github.com/Shopify/ruby-dev-exp-issues/issues/748

### Type of Change

- [ ] Add RBI for a new gem
- [x] Other: 

### Changes

Add launch configuration to support debugging through VS Code.

### DIFF
--- a/gem/rbi-central.gemspec
+++ b/gem/rbi-central.gemspec
@@ -32,6 +32,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency("spoom", ">= 1.1.12")
   spec.add_dependency("tapioca", ">= 0.9.2")
   spec.add_dependency("thor", ">= 1.2.1")
+  spec.add_development_dependency("debug")
 
   spec.required_ruby_version = ">= 3.1"
 end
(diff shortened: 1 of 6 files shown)


andyw8:
Not sure why this doesn't work... the command in the output looks correct:

`rdbg --command --open --stop-at-load -- bin/test /Users/andyw8/src/github.com/Shopify/rbi-central/gem/test/rbi-central/gem_test.rb`

But it fails with `Couldn't start debug session. The debuggee process exited with code 0` as if it's skipping any breakpoints that I add through the UI.

andyw8:
(FYI @st0012, I'll keep investigating but maybe you can see something obvious).

gem/rbi-central.gemspec
```diff
@@ -32,6 +32,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency("spoom", ">= 1.1.12")
   spec.add_dependency("tapioca", ">= 0.9.2")
   spec.add_dependency("thor", ">= 1.2.1")
+  spec.add_development_dependency("debug")
```
Morriar, line 35: Same, I think this only makes sense inside `gem/`

andyw8: Do you mean in `gem/Gemfile` rather than the `gemspec`?

Morriar: My bad I didn't read the filename properly 🤦 

Here's good 👍

PR #148, opened 2023-02-23
Add ActiveRecord callback signatures

paracycle:
These callback methods execute their blocks in the context of the model instance. This means that without these signatures, users will have to keep using `T.bind` inside the block to get the correct type for the binding of the block.

Ref: https://github.com/sorbet/sorbet/issues/6698

### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem name: activerecord
* Gem version: *


paracycle:
Tested this change in Core and it has only created 4 errors that are easily fixable (2 of them being messages to remove `T.bind` calls that are now redundant).

KaanOzkan (approved):

PR #149, opened 2023-02-25
Add signatures for Action Controller content_security_policy and permissions_policy

bdewater:
The block for content_security_policy [should have `bind(T.attached_class)` I believe](https://github.com/rails/rails/blob/93b1d3998cf5ce1e0cc91ebb0e6e65993bcec54a/actionpack/lib/action_controller/metal/content_security_policy.rb#L43) but that's not possible with modules at the moment.

### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other:

### DIFF
--- a/rbi/annotations/actionpack.rbi
+++ b/rbi/annotations/actionpack.rbi
@@ -4,6 +4,17 @@ class ActionController::API
   MODULES = T.let(T.unsafe(nil), T::Array[T.untyped])
 end
 
+module ActionController::ContentSecurityPolicy::ClassMethods
+  sig do
+    params(
+      enabled: T.untyped,
+      options: T.untyped,
+      block: T.nilable(T.proc.params(policy: ActionDispatch::ContentSecurityPolicy).void)
+    ).void
+  end
+  def content_security_policy(enabled = true, **options, &block); end
+end
+
 module ActionController::Flash::ClassMethods
   sig { params(types: Symbol).void }
   def add_flash_types(*types); end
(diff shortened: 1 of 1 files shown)


KaanOzkan (approved):

KaanOzkan (approved):
Sorry, forgot to merge this one. Added the `T.bind` like you said. Thanks.

rbi/annotations/actionpack.rbi
```diff
@@ -4,6 +4,23 @@ class ActionController::API
...
+  # @shim: required to declare T.attached_class
+  extend T::Generic
+
+  has_attached_class!(:out)
+
+  sig do
+    params(
+      enabled: T.untyped,
+      options: T.untyped,
+      block: T.nilable(T.proc.bind(T.attached_class).params(policy: ActionDispatch::ContentSecurityPolicy).void)
```
Morriar: I think we need to split the block binding by Rails version here. `content_security_policy` exists since Rails 5.2, but Rails used `yield policy` until Rails 7.1, so the block's `self` stayed as the controller class for 5.2, 6.x, and 7.0. The controller-instance binding only matches Rails 7.1+ where Rails switched to `instance_exec(policy, &block)`.

Could we keep the unbound proc for `< 7.1.0` and only use `bind(T.attached_class)` for `>= 7.1.0`? Same issue applies to `permissions_policy` below.

```ruby
# @version >= 5.2.0, < 7.1.0
block: T.nilable(T.proc.params(policy: ActionDispatch::ContentSecurityPolicy).void)

# @version >= 7.1.0
block: T.nilable(T.proc.bind(T.attached_class).params(policy: ActionDispatch::ContentSecurityPolicy).void)
```

PR #150, opened 2023-02-27
Add annotations for ActiveModel::Errors

bdewater:
### Type of Change

- [x] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem name: Active Model
* Gem version: 7.0.4.1
* Gem source: https://github.com/rails/rails/blob/v7.0.4.1/activemodel/lib/active_model/errors.rb
* Gem API doc: https://api.rubyonrails.org/classes/ActiveModel/Errors.html


bdewater:
We've typically dealt with this problem using `!`-suffixed methods that returns a non-nilable value.

An example from our app is `current_user` (from Devise) we know in a lot of places cannot be `nil` because of a controller `before_action` redirecting to a sign-in page, so we introduced `current_user!` to not deal with `T.must` wrapping everywhere.

KaanOzkan:
Thanks Bart 🙂 . I'm inclined to accept this annotation now. Next week I'll test a migration process in core and then we can merge.

As soon as the annotation is merged users can pull it in so we need to be ready for that and we can only turn off annotations on a per repo basis at the moment.

bdewater:
Ah that's new (and neat :)) - understand the concern!

rbi/annotations/activemodel.rbi
```diff
@@ -0,0 +1,85 @@
...
+  sig { params(options: T.untyped).returns(T::Hash[T.untyped, T.untyped]) }
+  def as_json(options = nil); end
+
+  sig { returns(T::Array[Symbol]) }
+  def attribute_names; end
+
+  sig { params(attribute: T.any(Symbol, String), type: T.untyped, options: T.untyped).returns(T.nilable(T::Array[String])) }
+  def delete(attribute, type = nil, **options); end
+
+  sig { returns(T::Hash[Symbol, T::Hash[Symbol, Symbol]]) }
```
KaanOzkan, line 24:
```suggestion
  sig { returns(T::Hash[Symbol, T::Array[T::Hash[Symbol, Symbol]]]) }
```
https://github.com/rails/rails/blob/8015c2c2cf5c8718449677570f372ceb01318a32/activemodel/lib/active_model/errors.rb#L250-L258

KaanOzkan (approved):

PR #151, opened 2023-03-09
Add a few methods and classes to the Stripe RBI

rafaelfranca:
Those methods are based in our codebase usage, and are in no way the entire list of existing objects and methods in the Stripe API.

### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem name: stripe
* Gem version: 8.3.0
* Gem source:https://github.com/stripe/stripe-ruby/tree/v8.3.0
* Gem API doc: https://stripe.com/docs/api
* Tapioca version: any
* Sorbet version: any

### DIFF
--- a/rbi/annotations/stripe.rbi
+++ b/rbi/annotations/stripe.rbi
@@ -1,7 +1,10 @@
 # typed: strict
 
 class Stripe::APIResource < Stripe::StripeObject
-  Elem = type_member {{fixed: T.untyped}}
+  Elem = type_member { { fixed: T.untyped } }
+
+  sig { params(id: T.any(String, T::Hash[Symbol, T.untyped]), opts: T.nilable(T::Hash[Symbol, T.untyped])).returns(Stripe::APIResource) }
+  def self.retrieve(id, opts = nil); end
 
   sig { returns(Stripe::APIResource) }
   def refresh; end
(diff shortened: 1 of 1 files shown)


rbi/annotations/stripe.rbi
```diff
@@ -18,33 +21,40 @@ class Stripe::APIResource < Stripe::StripeObject
...
+  sig { params(id: T.any(String, T::Hash[Symbol, T.untyped]), opts: T.nilable(T::Hash[Symbol, T.untyped])).returns(Stripe::ApplicationFee) }
+  def self.retrieve(id, opts = nil); end
 
   # @method_missing: from StripeObject
   sig { returns(String) }
   def id; end
+
+  # @method_missing: from StripeObject
+  sig { returns(Stripe::ListObject) }
+  def refunds; end
```
KaanOzkan, line 38: I couldn't tell where `refunds` and `reversals` are coming from. Loading [triggers](https://github.com/stripe/stripe-ruby/blob/e4e6bb54fdb53aef1edf8e92a14508abc61a3fa1/lib/stripe/resources/application_fee.rb#L11) `nested_resource_class_methods`. Which seems to [define](https://github.com/stripe/stripe-ruby/blob/e4e6bb54fdb53aef1edf8e92a14508abc61a3fa1/lib/stripe/api_operations/nested_resource.rb#L38) the methods prefixed with operations and a [method suffixed](https://github.com/stripe/stripe-ruby/blob/e4e6bb54fdb53aef1edf8e92a14508abc61a3fa1/lib/stripe/api_operations/nested_resource.rb#L22) with `_url`.

PR #153, opened 2023-03-09
Add annotations for shopify-money gem

olivier-thatch:
### Type of Change

- [x] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem name: shopify-money
* Gem version: 1.1.2
* Gem source: https://github.com/Shopify/money
* Gem API doc: 
* Tapioca version: Tapioca v0.10.5
* Sorbet version: Sorbet typechecker 0.5.10601 git 09bae4a8a0da2140f12443bc32a2725efc738eb7 debug_symbols=true clean=1

### DIFF
--- a/rbi/annotations/shopify-money.rbi
+++ b/rbi/annotations/shopify-money.rbi
@@ -7,6 +7,15 @@ class Money
   sig { returns(T.any(Money::Currency, Money::NullCurrency)) }
   attr_reader :currency
 
+  sig do
+    params(
+      value: T.nilable(T.any(Money, Numeric, String)),
+      currency: T.nilable(T.any(Money::Currency, Money::NullCurrency, String)),
+    )
+      .void
+  end
+  def initialize(value, currency); end
+
   # @method_missing: delegated to BigDecimal
   sig { params(args: T.untyped, _arg1: T.untyped, block: T.nilable(T.proc.void)).returns(T::Boolean) }
   def zero?(*args, **_arg1, &block); end
@@ -76,15 +85,6 @@ class Money
     end
     def with_currency(new_currency, &block); end
   end
-
-  sig do
-    params(
-      value: T.nilable(T.any(Money, Numeric, String)),
-      currency: T.nilable(T.any(Money::Currency, Money::NullCurrency, String)),
-    )
-      .void
-  end
-  def initialize(value, currency); end
 
   sig { params(format: Symbol).returns(Integer) }
   def subunits(format: :iso4217); end
--- a/rbi/annotations/shopify-money.rbi
+++ b/rbi/annotations/shopify-money.rbi
@@ -1,0 +1,51 @@
+# typed: strong
+
+class Money
+  sig { params(other: T.untyped).returns(Money) }
+  def +(other); end
+
+  sig { params(other: T.untyped).returns(Money) }
+  def -(other); end
+
+  sig { params(numeric: Numeric).returns(Money) }
+  def *(numeric); end
+
+  sig { params(numeric: Numeric).returns(T.noreturn) }
+  def /(numeric); end
+
+  sig { returns(String) }
+  def inspect; end
+
+  sig { params(other: T.untyped).returns(T::Boolean) }
+  def ==(other); end
+
+  sig { params(other: T.untyped).returns(T::Boolean) }
+  def eql?(other); end
+
+  sig { returns(Money) }
+  def abs; end
+
+  sig { returns(Money) }
+  def floor; end
+
+  sig { params(ndigits: Integer).returns(Money) }
+  def round(ndigits); end
+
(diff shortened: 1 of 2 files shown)


rbi/annotations/shopify-money.rbi
```diff
@@ -0,0 +1,51 @@
+# typed: strong
+
+class Money
```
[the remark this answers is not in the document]
olivier-thatch: `initialize` is slightly annoying to type because it will accept and try to coerce a lot of different types for both arguments, but I'll give it a try anyway.

The other methods are just an oversight, I'll add them too!

rbi/annotations/shopify-money.rbi
```diff
@@ -0,0 +1,173 @@
...
+  sig { returns(BigDecimal) }
+  attr_reader :value
+
+  sig { returns(T.any(Money::Currency, Money::NullCurrency)) }
+  attr_reader :currency
```
olivier-thatch, line 8: Annoyingly, `Money::NullCurrency` is not a subclass of `Money::Currency`, so we have to use a union.

rbi/annotations/shopify-money.rbi
```diff
@@ -0,0 +1,173 @@
...
+  sig { params(args: T.untyped, _arg1: T.untyped, block: T.nilable(T.proc.void)).returns(Integer) }
+  def to_i(*args, **_arg1, &block); end
+
+  # @method_missing: delegated to BigDecimal
+  sig { params(args: T.untyped, _arg1: T.untyped, block: T.nilable(T.proc.void)).returns(Float) }
+  def to_f(*args, **_arg1, &block); end
+
+  # @method_missing: delegated to BigDecimal
+  sig { params(args: T.untyped, _arg1: T.untyped, block: T.nilable(T.proc.void)).returns(Integer) }
+  def hash(*args, **_arg1, &block); end
```
olivier-thatch, line 36: The `(*args, **_arg1, &block)` arguments are necessary to appease Sorbet, because that's the signature it generates for methods delegated via `def_delegators`.

rbi/annotations/shopify-money.rbi
```diff
@@ -0,0 +1,173 @@
...
+    sig { params(currency: T.nilable(T.any(Money::Currency, Money::NullCurrency, String))).void }
+    def current_currency=(currency); end
+
+    sig do
+      params(
+        new_currency: T.nilable(T.any(Money::Currency, Money::NullCurrency, String)),
+        block: T.nilable(T.proc.void),
+      )
+        .void
+    end
```
KaanOzkan, line 85:
```suggestion
    sig do
      params(
        new_currency: T.nilable(T.any(Money::Currency, Money::NullCurrency, String)),
        block: T.nilable(T.proc.returns(T.untyped)),
      )
      .returns(T.untyped)
    end
```

Looks like the method yields to the block so `void` won't work here. https://github.com/Shopify/money/blob/2ebdb97174e743829542b3dafbf6cca42512d084/lib/money/money.rb#L74

rbi/annotations/shopify-money.rbi
```diff
@@ -0,0 +1,173 @@
...
+  class << self
+    sig { params(block: T.nilable(T.proc.params(config: Money::Config).void)).void }
+    def configure(&block); end
+
+    sig do
+      params(
+        value: T.nilable(T.any(Money, Numeric, String)),
+        currency: T.nilable(T.any(Money::Currency, Money::NullCurrency, String)),
+      )
+        .returns(Money)
```
KaanOzkan, line 56: This is now causing errors when `Money` instances are multiplied with integers, `5 * Money.new` as the second argument isn't an Integer https://github.com/sorbet/sorbet/blob/dd4602e030832c930d0b413f31734a83708448f6/rbi/core/integer.rbi#L59.

olivier-thatch: Ah, yeah, that's a bummer. I can't think of an easy fix though -- we'd need to be able to overload the signatures for `Integer#*`, `Float#*`, etc. with an additional signature, and I think Sorbet only supports overloading for its own stdlib types.

Maybe this is an acceptable limitation? It's easily fixed by swapping the arguments so that `Money#*` is called instead. It would be nice if this could be made more obvious but I'm not sure how.

Open to any and all suggestions for improving this :)

KaanOzkan: Yeah we can't represent the [current implementation](https://github.com/Shopify/money/blob/2ebdb97174e743829542b3dafbf6cca42512d084/lib/money/money.rb#L170-L194) without overloading that signature.

I'm okay with swapping the arguments solution. Only problem is it's hard to communicate this moving forward.

KaanOzkan (approved):

PR #154, opened 2023-03-09
More actionpack annotation changes

rafaelfranca:
### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other: 

### Changes

Each commit explain why the change.


rbi/annotations/actionpack.rbi
```diff
@@ -166,8 +166,8 @@ class ActionController::Parameters
   sig { params(keys: T.any(String, Symbol)).returns(ActionController::Parameters) }
   def slice(*keys); end
 
-  sig { returns(ActiveSupport::HashWithIndifferentAccess) }
-  def to_h; end
+  sig { params(block: T.nilable(T.proc.returns(T.untyped))).returns(ActiveSupport::HashWithIndifferentAccess) }
+  def to_h(&block); end
```
paracycle, line 170: Hmm, what does the block do in this case? I can't make the block get triggered:
```ruby
p = ActionController::Parameters.new({ a: { b: 1 } })
# => #<ActionController::Parameters {"a"=>{"b"=>1}} permitted: false>
p.permit(:a).to_h { |*args| p args.inspect }
# => {}
p.permit(a: {}).to_h { |*args| p args.inspect }
# => {"a"=>{"b"=>1}}
```

paracycle: I can't recreate that but if that is the case then maybe:
```suggestion
  sig { params(block: T.nilable(T.proc.params(arg: T.untyped).returns(T.untyped))).returns(ActiveSupport::HashWithIndifferentAccess) }
  def to_h(&block); end
```
might be better.

paracycle: OK, I guess that got added on Rails main. But, in this case the block actually accepts two arguments, but I guess there could be cases where it accepts a single argument too. Hmm.

paracycle: In that case, I think it is best to type the `block` as just a `T.nilable(Proc)` since that would not assert that the arities match. See the example [here](https://sorbet.run/?arg=--enable-experimental-requires-ancestor#%23%20typed%3A%20true%0A%0Aextend%20T%3A%3ASig%0A%0Asig%20%7B%20params%28block%3A%20T.nilable%28Proc%29%29.returns%28Hash%29%20%7D%0Adef%20to_h%28%26block%29%0A%20%20%7B%7D%0Aend%0A%0Asig%20%7B%20params%28block%3A%20T.nilable%28T.proc.returns%28T.untyped%29%29%29.returns%28Hash%29%20%7D%0Adef%20broken_to_h%28%26block%29%0A%20%20%7B%7D%0Aend%0A%0AT.reveal_type%20to_h%20%0A%0AT.reveal_type%20broken_to_h%0A%0AT.reveal_type%28to_h%20%7B%20%7Ck%7C%20k%20*%202%20%7D%29%0A%0AT.reveal_type%28broken_to_h%20%7B%20%7Ck%7C%20k%20*%202%20%7D%29).

```suggestion
  sig { params(block: T.nilable(Proc)).returns(ActiveSupport::HashWithIndifferentAccess) }
  def to_h(&block); end
```

PR #161, opened 2023-03-29
TimeWithZone calculations

randyn:
Include calculations that time has since `ActiveSupport::TimeWithZone` has them too https://api.rubyonrails.org/classes/ActiveSupport/TimeWithZone.html

### DIFF
--- a/rbi/annotations/activesupport.rbi
+++ b/rbi/annotations/activesupport.rbi
@@ -41,7 +41,9 @@ end
 
 class ActiveSupport::TimeWithZone
   include ::Comparable
+  # @shim: Methods on ActiveSupport::TimeWithZone are delegated to `Time` using `method_missing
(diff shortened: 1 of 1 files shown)


KaanOzkan (commented):
This looks good but can we actually add more while we're at it? This was a similar PR https://github.com/Shopify/rbi-central/pull/143 and since then I wanted to include `Time` methods under `TimeWithZone`.

We could start by including all of these mixins https://github.com/Shopify/shopify/blob/676266fa3989f4f067fcefbd914f0bbe4f62b1fc/sorbet/rbi/gems/activesupport%407.1.0.alpha-3025eaa09154f5f031f5c2e0e5b59553ff5c902b.rbi#L20048-L20051

randyn:
@KaanOzkan Sure thing, I added the other mixins and signed the CLA

PR #165, opened 2023-05-27
Add block param to Array#extract!

### LINKED ISSUE #164: ActiveSupport `Array#extract!` missing block param
### Problem

ActiveSupport `Array#extract!` [accepts an optional block arg](https://api.rubyonrails.org/v7.0.5/classes/Array.html#method-i-extract-21) but the RBI does not include the block argument:

https://github.com/Shopify/rbi-central/blob/main/rbi/annotations/activesupport.rbi#L145-L146

### Context

* Gem name: activesupport
* Gem version: 6.1.7.3
* Gem source: https://github.com/rails/rails/blob/main/activesupport/lib/active_support/core_ext/array/extract.rb
* Gem API doc: [Latest Version](https://api.rubyonrails.org/v7.0.5/classes/Array.html#method-i-extract-21) . The signature of this method [has never changed](https://github.com/rails/rails/commit/77b0126054178cb74c2b5bb17979462095144b7c#diff-69e4fa602a1f47160db003903d0de0671658b11a05dfe2883afe2988d429a032)
* Tapioca version: 0.11.6
* Sorbet version: 0.5.10845

bradleybuda:
Fix the signature by adding the missing parameter and improve it to include the block, its type, and the method return type

### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other: 

Fixes #164

### DIFF
--- a/rbi/annotations/activesupport.rbi
+++ b/rbi/annotations/activesupport.rbi
@@ -142,8 +142,11 @@ class Array
   sig { params(object: T.untyped).returns(T::Array[T.untyped]) }
   def self.wrap(object); end
 
-  sig { returns(T.untyped) }
-  def extract!; end
+  sig do
+    params(block: T.nilable(T.proc.params(element: Elem).returns(T.untyped)))
+      .returns(T::Array[T::Array[Elem]])
+  end
+  def extract!(&block); end
 
   sig { returns(ActiveSupport::ArrayInquirer) }
   def inquiry; end
--- a/rbi/annotations/activesupport.rbi
+++ b/rbi/annotations/activesupport.rbi
@@ -144,7 +144,7 @@ class Array
 
   sig do
     params(block: T.nilable(T.proc.params(element: Elem).returns(T.untyped)))
-      .returns(T::Array[T::Array[Elem]])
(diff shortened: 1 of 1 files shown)


bradleybuda:
I have signed the CLA!

rbi/annotations/activesupport.rbi
```diff
@@ -142,8 +142,11 @@ class Array
   sig { params(object: T.untyped).returns(T::Array[T.untyped]) }
   def self.wrap(object); end
 
-  sig { returns(T.untyped) }
-  def extract!; end
+  sig do
+    params(block: T.nilable(T.proc.params(element: Elem).returns(T.untyped)))
+      .returns(T::Array[T::Array[Elem]])
```
vinistock, line 147: I think adding the block is fine. I'm not so sure about this return though. I don't think this will work for invocations of `extract!` without a block, which [return an enumerator](https://github.com/rails/rails/blob/7d1c4dd7be1b51e2dcbb3614b68dc98c2926d28c/activesupport/lib/active_support/core_ext/array/extract.rb#L11).

Also, why is the return a nested array rather a simple one?

Issue #137, opened 2022-12-14
Add signatures for all Active Support extension methods on built-in types


egiurleo:
You can find the extensions within [this folder in the Rails repo](https://github.com/rails/rails/tree/main/activesupport/lib/active_support/core_ext).

- [ ] Array
- [ ] BigDecimal
- [ ] Class
- [ ] Date
- [ ] DateTime
- [ ] Digest
- [ ] Enumerable
- [ ] File
- [ ] Hash
- [ ] Integer
- [ ] Kernel
- [ ] LoadError
- [ ] Module
- [ ] NameError
- [ ] Numeric
- [ ] Object
- [ ] Pathname
- [ ] Range
- [ ] Regexp
- [ ] SecureRandom
- [ ] String
- [ ] Symbol
- [ ] Time
- [ ] URI

PR #166, opened 2023-05-30
Add duration methods to Numeric

grncdr:
See: #137

### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem name: activesupport
* Gem version: 
* Gem source:
  https://github.com/rails/rails/blob/main/activesupport/lib/active_support/core_ext/numeric/time.rb#L9
  https://github.com/rails/rails/blob/main/activesupport/lib/active_support/core_ext/integer/time.rb
* Gem API doc: https://api.rubyonrails.org/classes/Numeric.html#method-i-seconds (similar for other methods).
* Tapioca version: 0.11.6
* Sorbet version:

### DIFF
--- a/index.json
+++ b/index.json
@@ -34,6 +34,7 @@
       "active_support",
       "active_support/core_ext/array",
       "active_support/core_ext/numeric/time",
+      "active_support/core_ext/integer/time",
       "active_support/test_case"
     ]
   },
--- a/index.json
+++ b/index.json
@@ -33,6 +33,7 @@
     "requires": [
       "active_support",
       "active_support/core_ext/array",
+      "active_support/core_ext/numeric/time",
       "active_support/test_case"
     ]
   },
--- a/rbi/annotations/activesupport.rbi
+++ b/rbi/annotations/activesupport.rbi
@@ -59,7 +59,7 @@ class Hash
   def extractable_options?; end
 end
 
-class Integer
+class Numeric
   sig { returns(ActiveSupport::Duration) }
   def seconds; end
   sig { returns(ActiveSupport::Duration) }
--- a/rbi/annotations/activesupport.rbi
+++ b/rbi/annotations/activesupport.rbi
@@ -88,6 +88,12 @@ class Numeric
   def years; end
   sig { returns(ActiveSupport::Duration) }
   def year; end
+  sig { returns(ActiveSupport::Duration) }
+  def fortnight; end
+  sig { returns(ActiveSupport::Duration) }
+  def fortnights; end
+  sig { returns(Integer) }
+  def in_milliseconds; end
 end
 
 class Array
--- a/rbi/annotations/activesupport.rbi
+++ b/rbi/annotations/activesupport.rbi
@@ -59,6 +59,37 @@ class Hash
   def extractable_options?; end
 end
 
+class Integer
+  sig { returns(ActiveSupport::Duration) }
+  def seconds; end
+  sig { returns(ActiveSupport::Duration) }
+  def second; end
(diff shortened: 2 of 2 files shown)


andyw8:
I expect it's not commonly used, but let's include `fortnight` and `fortnights` for completeness?

And also `in_milliseconds`.

KaanOzkan (commented):
I think adding `active_support/core_ext/numeric/time` to [here](https://github.com/Shopify/rbi-central/blob/0eae2eb4d65208500c5502d709cfa026d3cfaabf/index.json#L33-L36) will fix CI

grncdr:
Signed the CLA

KaanOzkan (commented):
I'm conflicted about this PR because it'll cause a lot of typecheck errors due to the [delegation](https://github.com/rails/rails/blob/fdad62b23079ce1b90763cb5fa59321e7ac8b581/activesupport/lib/active_support/duration.rb#L508-L510) happening in `ActiveSupport::Duration`. I'm not sure if it'll bring in enough value. Applications are using `Duration` in place of `Numeric`.

For example, a common issue now would be while doing operations:
`Time.now - 1.second` errors with `Expected Time but found ActiveSupport::Duration for argument arg0` due to [sorbet's internal RBIs](https://github.com/sorbet/sorbet/blob/544f173782921ded834d925c5b86084326be244d/rbi/core/time.rbi#L368).

`ActiveSupport::Duration` is the correct type for `1.second` however it can be used in place of `Numeric` and it's hard to tell Sorbet that. We'll have to be able to override the internal RBI. Looks like `sorbet-typed` went [this route](https://github.com/sorbet/sorbet-typed/blob/440aa126c55570ad35cb5a5d4fe4a392f23d65d3/lib/activesupport/all/activesupport.rbi#L1442-L1443) before but I'm not sure how it was working for the above example.

KaanOzkan:
> It looks like there were overrides: https://github.com/sorbet/sorbet-typed/blob/440aa126c55570ad35cb5a5d4fe4a392f23d65d3/lib/activesupport/all/activesupport.rbi#L1252-L1266

Yes but I'm unable to make them work in our monolith. The overrides aren't picked up by Sorbet. I will be discussing this with Stripe and see if `sorbet-typed` is doing something we are not. Lmk if you have other findings in your application.

Even with the overrides I'm not convinced that typing these as `Duration` has practical typing benefits since we can't represent that delegation yet 😕

> It's also annoying to not be able to write 1.second.ago
Yeah currently the return type for `second` is coming from Tapioca generated gem RBIs for `activesupport` which is `T.untyped`. That's how we're able to write `1.second.ago`. I'd love to represent this delegation and add type safety to those call sites.

KaanOzkan:
Not really. We can't add `Duration` to sorbet's default RBIs so we'd have to return a `Numeric` from the methods in this PR which isn't correct.

ipvalverde:
Any updates on this one? We've been getting some annoying behaviour when we have something like the following:

```ruby
one_day_ago = Time.now.utc - 1.day
# one_day_ago is inferred to be a `Float`
```

KaanOzkan:
I was able to make overloads for stdlib methods like `Time#-` work so we'll able to cover most of the practical use cases by adding annotations for each of those methods in this file. However, there haven't been any changes to Sorbet regarding delegation so we still can't represent `ActiveSupport::Duration` accurately, i.e. `1.second.digits` will fail with a method `digits` does not exist o n`ActiveSupport::Duration` error. So when using methods that are called on the underlying type (`Integer` in this case) one would still need `T.cast`.

KaanOzkan:
Closing due to difficulty and prioritization.

PR #172, opened 2023-08-03
Add sidekiq's perform_bulk method signature

leeourand:
### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem name: sidekiq 
* Gem version: >= 6.5 
* Gem source: https://github.com/sidekiq/sidekiq/blob/352df475fceec348a543216f2f1fdddc047e8cc6/lib/sidekiq/job.rb#L240-L243 
* Gem API doc: https://www.rubydoc.info/gems/sidekiq/Sidekiq/Job/ClassMethods#perform_bulk-instance_method
* Tapioca version: 0.11.8
* Sorbet version: 0.5.10914

### DIFF
--- a/rbi/annotations/sidekiq.rbi
+++ b/rbi/annotations/sidekiq.rbi
@@ -75,6 +75,9 @@ module Sidekiq::Job::ClassMethods
 
   sig { params(interval: T.untyped, args: T.untyped).returns(String) }
   def perform_in(interval, *args); end
+
+  sig { params(args: T.untyped, kwargs: T.untyped).returns(String) }
+  def perform_bulk(*args, **kwargs); end
 end
 
 class Sidekiq::WorkSet
--- a/rbi/annotations/sidekiq.rbi
+++ b/rbi/annotations/sidekiq.rbi
@@ -76,7 +76,7 @@ module Sidekiq::Job::ClassMethods
   sig { params(interval: T.untyped, args: T.untyped).returns(String) }
   def perform_in(interval, *args); end
 
-  sig { params(args: T.untyped, kwargs: T.untyped).returns(String) }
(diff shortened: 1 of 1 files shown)


leeourand:
I have signed the CLA!

rbi/annotations/sidekiq.rbi
```diff
@@ -75,6 +75,9 @@ module Sidekiq::Job::ClassMethods
 
   sig { params(interval: T.untyped, args: T.untyped).returns(String) }
   def perform_in(interval, *args); end
+
+  sig { params(args: T.untyped, kwargs: T.untyped).returns(String) }
+  def perform_bulk(*args, **kwargs); end
```
vinistock, line 80: The `perform_bulk` method in `Job` indeed accepts [**kwargs](https://github.com/sidekiq/sidekiq/blob/352df475fceec348a543216f2f1fdddc047e8cc6/lib/sidekiq/job.rb#L315), but under the hood it invokes `Setter#perform_bulk`, which only accepts [batch_size](https://github.com/sidekiq/sidekiq/blob/352df475fceec348a543216f2f1fdddc047e8cc6/lib/sidekiq/job.rb#L240).

Can you actually invoke this method with any keyword arguments that aren't `batch_size`? If not, maybe it's worth making it explicit.

vinistock: This is actually telling Sorbet to accept keyword arguments and that each one of them will be a `{ batch_size: Integer }` shape. To get it to accept only `batch_size`, you need to do

```ruby
sig { params(args: T.untyped, batch_size: Integer).returns(String) }
def perform_bulk(*args, batch_size: 1000); end
```

PR #180, opened 2023-10-21
add basic devise annotation

fcheung:
### Type of Change

- [ x ] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem name: Devise
* Gem version: 4.9.3
* Gem source: https://github.com/heartcombo/devise
* Tapioca version: 0.11.9
* Sorbet version: 0.5.11074

This adds some very basic annotations for devise. Prior to tapioca 0.11.8 tapioca would generate rbis for Devise::SessionsController, Devise::RegistrationsController etc. (so that it could define helper related modules on them). As of tapioca 0.11.9 this is no longer the case, so subclasses of devise controllers cause a sorbet error whatever the type checking level requested on those files since the parent class is not known to sorbet.

I also added the actions for each controllers and the core devise controller methods

### DIFF
--- a/index.json
+++ b/index.json
@@ -49,6 +49,8 @@
   },
   "delayed_job": {
   },
+  "devise": {
+  },
   "elasticsearch-dsl": {
     "requires": [
       "elasticsearch/dsl"
--- a/rbi/annotations/devise.rbi
+++ b/rbi/annotations/devise.rbi
@@ -97,5 +97,23 @@ class Devise::SessionsController < DeviseController
   def sign_in_params; end
 end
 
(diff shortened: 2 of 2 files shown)


KaanOzkan:
If it's not possible to require the files that define these you could add a `@shim: description` tag on top of the constant definitions so that they aren't checked in runtime: https://github.com/Shopify/rbi-central/blob/f20fca307c2289233d9a578504f001591b13b02b/rbi/annotations/globalid.rbi#L4

KaanOzkan:
I haven't tested the signatures themselves because we don't use devise in the repos we own but I assume this was tested in your application.

PR #183, opened 2023-10-29
Annotate aasm instance method

bdewater:
### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem source: 
   - the method: https://github.com/aasm/aasm/blob/2e952ff2088082c9f00c95e3c1261afc826f737c/lib/aasm/aasm.rb#L67-L74
   - name parameter only being a String or Symbol: https://github.com/aasm/aasm/blob/2e952ff2088082c9f00c95e3c1261afc826f737c/lib/aasm/state_machine_store.rb#L66

### DIFF
--- a/rbi/annotations/aasm.rbi
+++ b/rbi/annotations/aasm.rbi
@@ -2,6 +2,9 @@
 
 module AASM
   mixes_in_class_methods(AASM::ClassMethods)
+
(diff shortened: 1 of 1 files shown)


vinistock:
Thank you for the contribution!

PR #184, opened 2023-10-30
Annotate remaining hardcoded blank?/present?/presence return values

bdewater:
After https://github.com/Shopify/rbi-central/pull/182#issuecomment-1783127074 decided to complete the set. `blank?` is defined and `present?` and `presence` and derived from there.

- [ ] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem name: Active Support
* Gem source:
  * https://github.com/rails/rails/blob/7-1-stable/activesupport/lib/active_support/core_ext/object/blank.rb
  * https://github.com/rails/rails/blob/7-1-stable/activesupport/lib/active_support/core_ext/date/blank.rb
  * https://github.com/rails/rails/blob/7-1-stable/activesupport/lib/active_support/core_ext/date_time/blank.rb
  * https://github.com/rails/rails/blob/7-1-stable/activesupport/lib/active_support/time_with_zone.rb#L504-L507

### DIFF
--- a/rbi/annotations/activesupport.rbi
+++ b/rbi/annotations/activesupport.rbi
@@ -44,6 +44,17 @@ class ActiveSupport::TimeWithZone
   include ::DateAndTime::Zones
   # @shim: Methods on ActiveSupport::TimeWithZone are delegated to `Time` using `method_missing
   include ::DateAndTime::Calculations
+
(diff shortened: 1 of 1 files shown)


vinistock (approved):
Thank you for the contribution!

PR #190, opened 2023-12-16
`starts_with?` and `ends_with?`

amomchilov:
### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem name: ActiveSupport
* Gem source: 
    * https://github.com/rails/rails/blob/main/activesupport/lib/active_support/core_ext/string/starts_ends_with.rb
    * https://github.com/rails/rails/blob/main/activesupport/lib/active_support/core_ext/symbol/starts_ends_with.rb
* Gem API doc: ...
* Tapioca version: `Tapioca v0.11.12`
* Sorbet version: `Sorbet typechecker 0.5.11142`


rbi/annotations/activesupport.rbi
```diff
@@ -259,3 +259,23 @@ class Time
   sig { returns(TrueClass) }
   def present?; end
 end
+
+class Symbol
+  # @shim: alias for `#start_with?`
```
Morriar, line 264: Does it really need to be marked as `@shim`?

```rb
require "active_support"
Symbol.instance_method(:starts_with?) # => #<UnboundMethod: Symbol#starts_with?(start_with?)(*)>
```

Morriar: To avoid errors and outdated RBIs, `rbi-central` CI will check that the method actually exist at runtime.

`@shim` is used to disable this check when the method doesn't actually exist and the RBI definition is used to shim the actual behavior to make typing possible: https://github.com/Shopify/rbi-central?tab=readme-ov-file#runtime-checks

In your case, I don't think you need it since the definition actually exists.

amomchilov: Oh neat, thanks for pointing me to that!

Removed in #208

PR #214, opened 2024-01-08
Add RBIs for all ActiveSupport String extensions

amomchilov:
### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem name: ActiveSupport
* Gem source: 
    * https://github.com/rails/rails/blob/main/activesupport/lib/active_support/core_ext/string
* Gem API doc: ...
* Tapioca version: `Tapioca v0.11.12`
* Sorbet version: `Sorbet typechecker 0.5.11142`

### DIFF
--- a/rbi/annotations/activesupport.rbi
+++ b/rbi/annotations/activesupport.rbi
@@ -271,6 +271,150 @@ class Symbol
 end
 
 class String
+  sig { returns(TrueClass) }
+  def acts_like_string?; end
+
+  sig { returns(String) }
+  def as_json; end
+
+  # Same as the sigs for `#[]`
+  # https://github.com/sorbet/sorbet/blob/40ad87b4dc7be23fa00c1369ac9f927053c68907/rbi/core/string.rbi#L270-L303
+  sig { params(arg0: Integer,  arg1: Integer).returns(T.nilable(String)) }
+  sig { params(arg0: T.any(T::Range[Integer], Regexp)).returns(T.nilable(String)) }
+  sig { params(arg0: Regexp, arg1: Integer).returns(T.nilable(String)) }
+  sig { params(arg0: Regexp, arg1: String).returns(T.nilable(String)) }
+  sig { params(arg0: String).returns(T.nilable(String)) }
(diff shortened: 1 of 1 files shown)


rbi/annotations/activesupport.rbi
```diff
@@ -271,6 +271,148 @@ class Symbol
...
+  sig { returns(ActiveSupport::Multibyte::Chars) }
+  def mb_chars; end
+
+  sig { params(separator: String, preserve_case: T::Boolean, locale: T.nilable(Symbol)).returns(String) }
+  def parameterize(separator: "-", preserve_case: false, locale: nil); end
+
+  sig { params(count: Integer, locale: T.nilable(String)).returns(String) }
+  def pluralize(count = nil, locale = :en); end
+
+  sig { params(patterns: T.any(String, Regexp)) }
```
andyw8, line 359: Should this always return a string?

https://api.rubyonrails.org/classes/String.html#method-i-remove

rbi/annotations/activesupport.rbi
```diff
@@ -271,6 +271,148 @@ class Symbol
...
+  sig { params(count: Integer, locale: T.nilable(String)).returns(String) }
+  def pluralize(count = nil, locale = :en); end
+
+  sig { params(patterns: T.any(String, Regexp)).returns(String) }
+  def remove(*patterns); end
+
+  sig { params(patterns: T.any(String, Regexp)).returns(String) }
+  def remove!(*patterns); end
+
+  sig { returns(T.untyped) }
```
andyw8, line 365: Should this return `T.nilable(T.any(Class, Module))`?

amomchilov: Nope! A constant can hold any kind of value, e.g.

```ruby
ABC = 123
"ABC".safe_constantize # => 123
```

KaanOzkan (approved):
@amomchilov Just a heads up the `activesupport` annotation is ignored in the tapioca config in core so you can't observe any errors. Sigs look OK to me, other applications can benefit from them and report back if there are issues.

PR #221, opened 2024-02-14
Annotations for Active Support error reporter

bdewater:
### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem name: Active Support
* Gem version: 7.1.3
* Gem source: https://github.com/rails/rails/blob/v7.1.3/activesupport/lib/active_support/error_reporter.rb
* Gem API doc: https://api.rubyonrails.org/v7.1.3/classes/ActiveSupport/ErrorReporter.html
* Tapioca version: 
* Sorbet version:

### DIFF
--- a/rbi/annotations/activesupport.rbi
+++ b/rbi/annotations/activesupport.rbi
@@ -427,3 +427,43 @@ class String
   sig { returns(String) }
   def upcase_first; end
 end
+
+class ActiveSupport::ErrorReporter
+  sig do
+    type_parameters(:Block, :Fallback)
+      .params(
+        error_classes: T.class_of(Exception),
+        severity: T.nilable(Symbol),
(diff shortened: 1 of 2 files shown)


ana-c-santos:
Think this might have introduced a bug. The definitions for the functions have optional arguments with a default value that references a constant that is not declared anywhere, causing problems when running Sorbet.

vinistock:
Are you talking about `DEFAULT_SOURCE`? Are you on the latest Rails version? Maybe the constant doesn't exist in older versions and that's the issue.

ana-c-santos:
> Are you talking about `DEFAULT_SOURCE`? Are you on the latest Rails version? Maybe the constant doesn't exist in older versions and that's the issue.

You are right, I am using 7.0.8 and this constant seems to have been introduced in 7.1

PR #222, opened 2024-02-15
Remove `DEFAULT_SOURCE` from Active Support RBI

vinistock:
In #221, we added some new signatures for new Rails methods. However, the `DEFAULT_SOURCE` constant doesn't exist on older versions of Rails, which results in a type checking error.

I put `nil` instead of a random string. It's not correct, but using a random string didn't feel right either. Let me know if you have a different preference.

### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other:

### DIFF
--- a/rbi/annotations/activesupport.rbi
+++ b/rbi/annotations/activesupport.rbi
@@ -441,7 +441,7 @@ class ActiveSupport::ErrorReporter
       )
       .returns(T.any(T.type_parameter(:Block), T.type_parameter(:Fallback)))
   end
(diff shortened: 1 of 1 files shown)


andyw8 (approved):

egiurleo (approved):

paracycle (commented):
I think it is a mistake to try to supply default values to RBIs, since there is almost no value to be gained from them. Moreover, a default value can be (a) referencing other parameters and (b) be as arbitrarily complex as possible (think `begin`/`end` blocks with lots of code in between).

The best practice is to type every default value as `T.unsafe(nil)` to show that we don't care about what the default value is, but we care that there _is_ a default value.

PR #225, opened 2024-02-26
Change ActiveSupport::ErrorReporter#report to return NilClass

bdewater:
Follow-up to https://github.com/Shopify/rbi-central/pull/221 

This is what it actually returns and makes error reporting at the end of a method that returns `T.nilable(MyModel)` easier - I don't have to put `nil` return values myself anymore.

### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem name: Active Support
* Gem version: 7.1.3
* Gem source: https://github.com/rails/rails/blob/v7.1.3/activesupport/lib/active_support/error_reporter.rb#L200

### DIFF
--- a/rbi/annotations/activesupport.rbi
+++ b/rbi/annotations/activesupport.rbi
@@ -463,7 +463,7 @@ class ActiveSupport::ErrorReporter
       severity: T.nilable(Symbol),
       context: T::Hash[Symbol, T.untyped],
       source: T.nilable(String),
-    ).void
+    ).returns(NilClass)
   end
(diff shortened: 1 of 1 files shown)


KaanOzkan:
I think Vini is right, looking at https://sorbet.org/docs/sigs#returns--void-annotating-return-types this falls under the description exactly. I'd prefer if this stayed as void to represent the intent. It's annoying given the usecases of report but it's safer. Although the fact that it returns `nil` does mean benefits of showing this intent is limited. Does it work if you do this diff in a shim specific to your project?

bdewater:
I can def work around it one way or the other, it just felt annoying enough to submit this PR :D

PR #262, opened 2024-07-15
Return TimeWithZone from Time.current

ipvalverde:
### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem name: `activesupport`
* Gem version: 
* Gem source: 
* Gem API doc: [Time.current](https://api.rubyonrails.org/classes/Time.html#method-c-current) and [Time.zone](https://api.rubyonrails.org/classes/Time.html#method-c-zone)
* Tapioca version: 
* Sorbet version: 

### Description

Fixes https://github.com/Shopify/rbi-central/issues/261

Although the documentation does indicate that [`Time.current` should be able to return a `Time` object](https://api.rubyonrails.org/classes/Time.html#method-c-current) this [is unlikely to happen in a Rails application](https://github.com/Shopify/rbi-central/issues/261#issuecomment-2218566655). This PR is to mitigate the need of extra code from consumers that might want to assign the return of `Time.current` to a `TimeWithZone`.

Ideally we should make the `TimeWithZone` to inherit from `Time`, since it hacks its way to be like a `Time`, but this was [failing CI](https://github.com/Shopify/rbi-central/actions/runs/9938221887/job/27450216515) and it can be in a different PR.

### DIFF
--- a/rbi/annotations/activesupport.rbi
+++ b/rbi/annotations/activesupport.rbi
@@ -267,9 +267,6 @@ class Time
 
   sig { returns(ActiveSupport::TimeZone) }
   def self.zone; end
-
-  sig { returns(T.any(ActiveSupport::TimeWithZone, ::Time)) }
-  def self.current; end
 end
 
 class Symbol


paracycle:
Active Support is not only used in Rails applications, so making the return type fit just a majority of Rails applications  is problematic in my opinion.

Since this is dynamic and the return value depends on the presence of `Time.zone`, I'd rather that we create a DSL compiler for this in Tapioca which would create the right signature for the application it is working against.

paracycle:
I just created a PR for the compiler, btw: https://github.com/Shopify/tapioca/pull/1962

amomchilov:
Shall we close this draft then?

PR #263, opened 2024-07-16
Bring `ActiveJob#discard_on` RBI in-line with `rescue_on` and `retry_on` methods

larouxn:
### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem name: activejob
* Gem version: N/A
* Gem source: https://github.com/rails/rails/tree/main/activejob
* Gem API doc: https://api.rubyonrails.org/classes/ActiveJob.html
* Tapioca version: 0.15.1 (latest as of July 216 2024)
* Sorbet version: 0.5.11481 (latest as of July 216 2024)

I'm trying to bring `discard_on` in-line with `rescue_on` and `retry_on` considering according to the Rails documentation they're all the same.

- https://api.rubyonrails.org/classes/ActiveJob/Exceptions/ClassMethods.html#method-i-discard_on
- https://api.rubyonrails.org/classes/ActiveJob/Exceptions/ClassMethods.html#method-i-retry_on
- https://api.rubyonrails.org/classes/ActiveSupport/Rescuable/ClassMethods.html#method-i-rescue_from

Additionally, would seemingly allow `discard_on` usage to avoid needing `T.unsafe` with the latest Sorbet provided the array of strings is frozen via `freeze` re: the following change.
- https://github.com/sorbet/sorbet/pull/7993


egiurleo (approved):

larouxn:
🎩 Just updated local ActiveJob RBI annotations file manually and indeed this appears to fix the discrepancy behavior between splatting `discard_on` and `rescue_on` + `retry_on` when using a frozen array of strings. Before `discard_on` needed `T.unsafe` and with this change it appears it no longer does. 🙌

egiurleo:
Thanks @larouxn!

PR #264, opened 2024-07-17
Add ErrorReporter unexpected method

bdewater:
### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem name: Active Support
* Gem version: 7.2 beta
* Gem source: https://github.com/rails/rails/blob/9098f535b2accef27c6c284dede119c936c44595/activesupport/lib/active_support/error_reporter.rb#L122-L154

### DIFF
--- a/rbi/annotations/activesupport.rbi
+++ b/rbi/annotations/activesupport.rbi
@@ -493,4 +493,15 @@ class ActiveSupport::ErrorReporter
     ).void
   end
   def report(error, handled: true, severity: T.unsafe(nil), context: T.unsafe(nil), source: T.unsafe(nil)); end
+
+  # @version >= 7.2.0.beta1
+  sig do
+    params(
+      error: T.any(Exception, String),
+      severity: T.nilable(Symbol),
+      context: T::Hash[Symbol, T.untyped],
+      source: T.nilable(String),
+    ).void
+  end
+  def unexpected(error, severity: T.unsafe(nil), context: T.unsafe(nil), source: T.unsafe(nil)); end
 end


rbi/annotations/activesupport.rbi
```diff
@@ -472,4 +472,15 @@ class ActiveSupport::ErrorReporter
     ).void
   end
   def report(error, handled: true, severity: T.unsafe(nil), context: T.unsafe(nil), source: T.unsafe(nil)); end
+
+  # @version >= 7.2.0.beta1
```
egiurleo, line 476: I think you win the award for the very first version annotation? 🥇

KaanOzkan: How do we want to handle runtime checks? I think CI will use the latest version however beta versions can only be specified in the Gemfile?

There's also the case of annotations with an incompatible version, `< 7.2.0beta1`. Should we just use `@method_missing` instead of dealing with this?

bdewater:
With Rails 7.2 out this now passes CI.

rbi/annotations/activesupport.rbi
KaanOzkan: I think this PR is fine thanks to https://github.com/Shopify/tapioca/pull/1585 but I'll leave it to @egiurleo before merging. We may wanna test versioning with apps that use lower Rails versions as we merge this.

egiurleo: Yeah let's merge this and I'll add a note to our meeting on Monday to talk about it.

egiurleo (approved):

PR #271, opened 2024-07-25
Fix incorrect usages of `T.nilable(X, Y)` in signatures

Morriar:
### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other

### Changes

In both `lhm` and `lhm-shopify` signatures contained calls to `T.nilable` with more than one type. This is an error in Ruby files and should be an error in RBI files (once https://github.com/sorbet/sorbet/issues/8054 is fixed).


amomchilov (approved):
Interesting, how did you stumble into these? Do they even typecheck correctly?

PR #275, opened 2024-08-05
Add Phonelib area code

bdewater:
This bit me when trying to use the area code with `delete_prefix`, which doesn't accept `nil`.

### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem name: Phonelib
* Gem version: 0.9.1
* Gem source: https://github.com/daddyz/phonelib/blob/466a328ec14d25e95542dc7eb70a358e051620bb/lib/phonelib/phone_formatter.rb#L96-L108

### DIFF
--- a/rbi/annotations/phonelib.rbi
+++ b/rbi/annotations/phonelib.rbi
@@ -30,6 +30,9 @@ module Phonelib
   end
 
   module PhoneFormatter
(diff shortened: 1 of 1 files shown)


egiurleo (approved):

egiurleo:
Thanks @bdewater!

PR #278, opened 2024-08-13
Add versioning to ErrorReporter methods and backfill 7.0 signatures

bdewater:
### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem name: Active Support
* Gem source:
  * https://github.com/rails/rails/blob/v7.0.0/activesupport/lib/active_support/error_reporter.rb
  * https://github.com/rails/rails/blob/v7.1.0.beta1/activesupport/lib/active_support/error_reporter.rb
* Gem API doc:
  * https://api.rubyonrails.org/v7.0.0/classes/ActiveSupport/ErrorReporter.html
  * https://api.rubyonrails.org/v7.1.0/classes/ActiveSupport/ErrorReporter.html

### DIFF
--- a/rbi/annotations/activesupport.rbi
+++ b/rbi/annotations/activesupport.rbi
@@ -456,6 +456,21 @@ class String
 end
 
 class ActiveSupport::ErrorReporter
+  # @version ~> 7.0.0
+  sig do
+    type_parameters(:Block, :Fallback)
+      .params(
+        error_class: T.class_of(Exception),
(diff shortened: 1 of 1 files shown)


egiurleo:
I think the issues preventing this from being merged were adequately addressed by @KaanOzkan in https://github.com/Shopify/rbi-central/pull/286. Woohoo!

egiurleo (approved):

PR #286, opened 2024-09-27
Filter versions while testing

KaanOzkan:
Preliminary support for testing annotations with versions. 

In a future improvement we should test all versions and not just one used in CI which should be the latest version: https://github.com/Shopify/rbi-central/issues/287


egiurleo (approved):

gem/lib/rbi-central/static/context.rb
```diff
@@ -78,6 +79,17 @@ def add_require(name)
           require "#{name}"
         RB
       end
+
+      private
+
+      sig { params(gem_name: String, annotations_file: String).returns(String) }
```
Morriar, line 85: I wonder if this should go in the parent class?

Morriar (approved):
Thanks!

PR #298, opened 2024-12-20
Move over Sorbet stdlib types for the `cmath` gem

Tom-Bonnike:
TBD

### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem name: 
* Gem version: 
* Gem source: 
* Gem API doc: 
* Tapioca version: 
* Sorbet version:

### DIFF
--- a/rbi/annotations/cmath.rbi
+++ b/rbi/annotations/cmath.rbi
@@ -1,0 +1,60 @@
+module CMath
+  include ::Math
+
+  sig { params(z: Numeric).returns(Float) }
+  def self.acos(z); end
+
+  sig { params(z: Numeric).returns(Float) }
+  def self.acosh(z); end
+
+  sig { params(z: Numeric).returns(Float) }
+  def self.asin(z); end
+
+  sig { params(z: Numeric).returns(Float) }
+  def self.asinh(z); end
+
+  sig { params(z: Numeric).returns(Float) }
+  def self.atan(z); end
+
+  sig { params(y: Numeric, x: Numeric).returns(Float) }
+  def self.atan2(y, x); end
+
+  sig { params(z: Numeric).returns(Float) }
+  def self.atanh(z); end
+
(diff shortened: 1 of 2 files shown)


rbi/annotations/cmath.rbi
```diff
@@ -0,0 +1,62 @@
+# typed: true
+
+module CMath
+  include ::Math
```
KaanOzkan, line 4: This is failing CI for now but the duplication should be fine as users upgrade their Sorbet version. I'd like to keep the CI green though. Is it okay if this PR waits until Sorbet removal is merged and we bump the Sorbet version used in tests?

Tom-Bonnike: I guess that makes sense — although we should aim for as little time as possible between the two merges, then.

Tom-Bonnike: Hi @KaanOzkan, the PR was just merged on Sorbet’s side, and a new release was published!

KaanOzkan: That gem (`rbi-central`) is only used in this project and it's not published. It's still good to bump it so that CI uses the correct sorbet version and doesn't error. We can't control the sorbet version from here but now that I'm taking a closer look I don't think we have to.

I'm not sure why CI is failing now (static check, rubocop failures I'll fix in `main`). Testing this annotation on new and old sorbet versions it works fine. Also tapioca generates the `include ::Math` so we don't need to have it in this annotation.

Tom-Bonnike: Removed that, and re-updated sorbet, I think it hadn’t updated to the necessary version somehow, hopefully CI will pass now

KaanOzkan (approved):
CI failures are expected. I tested the annotations with new and older versions of Sorbet, since signatures are identical they don't cause errors for Sorbet versions that have CMath definitions in its payload.

Issue #231, opened 2024-03-05
`ActionController::Parameters#require` signature needs adjustment


andyw8:
See https://github.com/Shopify/rbi-central/pull/224#pullrequestreview-1917467194

amomchilov:
@aprescott Do you know what the full possible set of returnable types is?

It should be `T.any(String, Integer, T::Array[T.any(String, Integer))` at a minimum, but do you know of any other possible cases?

Whatever that type is, that's what should be used as the return value of not just `#require`, but also `#[]`, which still [returns `T.untyped`](https://github.com/bdewater/rbi-central/blob/c8b5f1a749af856dc3d79fdaeb7b4d74b13f687c/rbi/annotations/actionpack.rbi#L111-L112)

> although interestingly it doesn't always fire runtime errors, as in cases like a `T::Array[X]` vs. `T::Array[Y]`.

That's a performance concession, Sorbet's runtime type-checking doesn't do deep recursive type-checks on Enumerables. The runtime type-checking calls [`#valid?`](https://github.com/sorbet/sorbet/blob/2d91ebab7b642b900e02233bd49f593ad355469d/gems/sorbet-runtime/lib/types/types/typed_array.rb#L20-L23), not [`#recursively_valid?`](https://github.com/sorbet/sorbet/blob/2d91ebab7b642b900e02233bd49f593ad355469d/gems/sorbet-runtime/lib/types/types/typed_enumerable.rb#L37)

aprescott:
> Do you know what the full possible set of returnable types is?

I'm not sure! It would be good to match whatever `Parameters` can possibly support. String or Integer or Array containers of those certainly sounds reasonable. Perhaps Rails source would help confirm. I don't really know much about the rbi-central project's approach here, but perhaps `T.anything` is an option if the full set of output types isn't clear? The true correct typing would of course beat `T.anything`.

> Sorbet's runtime type-checking doesn't do deep recursive type-checks on Enumerables

Found the docs explaining this right after my comment, realized my confusion!

amomchilov:
> String or Integer or Array containers of those certainly sounds reasonable.

Agreed, let's start there, and adjust it if people find that other values were possible.

@aprescott Would you be interested in opening your first RBI Central PR to contribute this Sig? :)

> I don't really know much about the rbi-central project's approach here, but perhaps `T.anything` is an option if the full set of output types isn't clear?

I like using `T.anything` in my own code, but it's not a good fit for RBI central, because:

[...]

PR #313, opened 2025-04-23
Fix ActionController::Parameters#require sigs

duelinmarkers:
as raised in #231. This is often a scalar, rather than nested parameters, but it can be anything.

### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem name: actionpack
* Gem version: at least >= 6
* Gem source: https://github.com/rails/rails/blob/3235827585d87661942c91bc81f64f56d710f0b2/actionpack/lib/action_controller/metal/strong_parameters.rb#L519
* Gem API doc: https://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-require
* Tapioca version: 0.16.2
* Sorbet version: 0.5.11139

### DIFF
--- a/rbi/annotations/actionpack.rbi
+++ b/rbi/annotations/actionpack.rbi
@@ -111,13 +111,13 @@ class ActionController::Parameters
   sig { params(key: T.any(String, Symbol)).returns(T.untyped) }
   def [](key); end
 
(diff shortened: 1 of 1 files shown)


duelinmarkers:
I have signed the CLA!

Morriar:
Thanks!

PR #338, opened 2025-09-24
Add signature for assert_changes method

ipvalverde:
### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem name: Rails / ActiveSupport
* Gem version: 
* Gem source: [link](https://github.com/rails/rails/blob/b0c813bc7b61c71dd21ee3a6c6210f6d14030f71/activesupport/lib/active_support/testing/assertions.rb#L195)
* Gem API doc: [link](https://api.rubyonrails.org/classes/ActiveSupport/Testing/Assertions.html#method-i-assert_changes)
* Tapioca version: 
* Sorbet version: 

This will bring two main benefits from typing:

#### Propagate block's return type

The return of the block will now be propagated to as the return type of the method, no longer "untyping" the block's return:

```ruby
create_sale_result = assert_changes -> { Sales.count } do
  CreateSale.perform
end

# create_sale_result will be typed in here
```

#### Type message argument as string

The second positional argument for the method is enforced to be a string. The assert_changes has the following signature ([docs](https://api.rubyonrails.org/classes/ActiveSupport/Testing/Assertions.html#method-i-assert_changes)):

```ruby
def assert_changes(expression, message = nil, from: UNTRACKED, to: UNTRACKED, &block); end
```

Unfortunately, the method implementation is too permissive and allow anything to go in, easily allowing misuses:

```ruby
# In this case, the assertion is only checking that the value of the expression `Sales.count` changes,
# it doesn't necessarily check the delta change to be `+784`. The second argument is just the string message
# for when it fails...
# If this fails, it will print: "784."
assert_changes -> { Sales.count }, +784 # <-- 🙈
  Sales.last.destroy!
end

# The correct usage with the second argument would be something like this:
assert_changes -> { Sales.count }, "Sales did not decrement as expected", from: initial_sales_count, to: expected_sales_count
  Sales.last.destroy!
end
```

Are there any concerns on restricting the second argument to be a string? In theory, any kind of object can be passed as the argument and it's string representation will be used to generate the error message.
I believe setting the message as `string` is a good trade off to have Sorbet picking up wrong usages of the assertion ([example](https://github.com/search?q=%2Fassert_changes%28%28%5B%5Cs%5Cr%5Cn%5D*%5C%28%3F-%3E%5B%5Cs%5Cr%5Cn%5D*%5C%7B%5B%5Cs%5Cr%5Cn%5D*%5B%5E%7D%5D%2B%5C%7D%5B%5Cs%5Cr%5Cn%5D*%2C%5B%5Cs%5Cr%5Cn%5D*%5B%5C%2B%5C-%5D%3F%28%5Cd%2B%29%29%29%2F+language%3Aruby&type=code))


rbi/annotations/activesupport.rbi
```diff
@@ -594,4 +594,17 @@ module ActiveSupport::Testing::Assertions
     type_parameters(:Block).params(block: T.proc.returns(T.type_parameter(:Block))).returns(T.type_parameter(:Block))
   end
   def assert_nothing_raised(&block); end
+
+  sig do
+    type_parameters(:TResult)
+      .params(
+        expression: T.untyped,
+        message: T.nilable(String),
```
KaanOzkan, line 602: Thanks for flagging this. I think we need to be accurate here instead of opinionated. Message can be anything that has `to_s`. We can't represent that perfectly but maybe we can do `Object`, if not `T.anything`.

ipvalverde: Yeah, I thought that would be the case. Will try to check in Rails if it makes sense to add a runtime validation for this message instead, will swap it to `T.anything`.

KaanOzkan (approved):

PR #361, opened 2026-01-02
Update Tapioca-generated RBIs 

amomchilov:
### Type of Change

We have a number of tapioca-generated RBIs that are quite stale. I updated these with `bin/tapioca gem --all`.


KaanOzkan:
Outdated, looks like we had `tapioca gem` runs since then.

PR #382, opened 2026-04-10
Add annotations for discard gem callbacks

TakuroOnoda:
### Description

Adds annotations for callback methods defined dynamically by `define_model_callbacks` in `Discard::Model`.

When a model includes `Discard::Model`, the following class methods are defined at runtime via `define_model_callbacks :discard` and `define_model_callbacks :undiscard`:

- `before_discard`, `around_discard`, `after_discard`
- `before_undiscard`, `around_undiscard`, `after_undiscard`

These cannot be captured by `tapioca gem` because they are generated dynamically inside the `included` block, so they are annotated with `@shim`.

---

### Type of Change

- [x] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem name: discard
* Gem version: 1.4.0
* Gem source: https://github.com/jhawthorn/discard/blob/master/lib/discard/model.rb#L21-L22
* Gem API doc: https://github.com/jhawthorn/discard#callbacks
* Tapioca version: 0.17.10
* Sorbet version: 0.6.12913

---

Thank you for taking the time to review this PR!

### DIFF
--- a/rbi/annotations/discard.rbi
+++ b/rbi/annotations/discard.rbi
@@ -1,0 +1,27 @@
+# typed: true
+
+module Discard::Model::ClassMethods
+  # @shim: defined dynamically via `define_model_callbacks :discard`
+  sig { params(args: T.untyped, block: T.nilable(T.proc.bind(T.untyped).void)).void }
+  def before_discard(*args, &block); end
+
+  # @shim: defined dynamically via `define_model_callbacks :discard`
+  sig { params(args: T.untyped, block: T.nilable(T.proc.bind(T.untyped).void)).void }
+  def around_discard(*args, &block); end
+
+  # @shim: defined dynamically via `define_model_callbacks :discard`
+  sig { params(args: T.untyped, block: T.nilable(T.proc.bind(T.untyped).void)).void }
+  def after_discard(*args, &block); end
+
+  # @shim: defined dynamically via `define_model_callbacks :undiscard`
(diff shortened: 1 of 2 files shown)


TakuroOnoda:
I have signed the CLA!

rbi/annotations/discard.rbi
```diff
@@ -0,0 +1,27 @@
+# typed: true
+
+module Discard::Model::ClassMethods
+  # @shim: defined dynamically via `define_model_callbacks :discard`
+  sig { params(args: T.untyped, block: T.nilable(T.proc.bind(T.untyped).void)).void }
+  def before_discard(*args, &block); end
```
KaanOzkan: Thanks, this looks good. But I think we should also include `**options` in the sig & parameter list. Inspecting the parameters using `Rails 8.1.3` I found that each method takes a double splat argument:
`before_discard: [[:rest, :args], [:keyrest, :options], [:block, :block]]`

So I think we need:
```rb
sig { params(args: T.untyped, options: T.untyped, block: T.nilable(T.proc.bind(T.untyped).void)).void }
def before_discard(*args, **options, &block); end
```

and so on for others.

TakuroOnoda: @KaanOzkan 
Thank you for the review!
Updated methods to include `**options` in both the sig and parameter list.

KaanOzkan (approved):

PR #388, opened 2026-04-20
Add signatures for Hash and ActiveSupport::HashWithIndifferentAccess

pyrkhanton:
Right now when call `#with_indifferent_access` on `Hash` or `ActiveSupport::HashWithIndifferentAccess` result is `T.untyped`.
Correct this behavior to return proper class - `ActiveSupport::HashWithIndifferentAccess`.
Solution is to add sig for `#with_indifferent_access` to return `ActiveSupport::HashWithIndifferentAccess`.

Similar behavior is met for some of instance methods of `ActiveSupport::HashWithIndifferentAccess`.
As example methods provided by `ActiveSupport::DeepMergeable`: `#deep_merge`, `#deep_merge!`, `#deep_merge?`.
Added signatures for them.
For `ActiveSupport::HashWithIndifferentAccess` methods returned result type should be `ActiveSupport::HashWithIndifferentAccess`.

If this would makes sense to do - it may be expanded to other `ActiveSupport::HashWithIndifferentAccess` methods (provide more similar signatures both for `Hash` and `ActiveSupport::HashWithIndifferentAccess` that are provided by `activesupport`)

### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem name: activesupport
* Gem version: 7.0.8.4
* Gem source: https://github.com/rails/rails/tree/v7.0.8.4/activesupport/lib/active_support
* Tapioca version: 0.18.0
* Sorbet version: 0.6.13055

### DIFF
--- a/rbi/annotations/activesupport.rbi
+++ b/rbi/annotations/activesupport.rbi
@@ -52,6 +52,7 @@ class ActiveSupport::HashWithIndifferentAccess < Hash
   end
   def deep_merge!(other_hash, &block); end
 
+  # @version >= 7.2
   sig { params(other: T.untyped).returns(T::Boolean) }
   def deep_merge?(other); end
 end
@@ -124,6 +125,9 @@ class Hash
   sig { returns(T.self_type) }
   def compact_blank; end
 
+  sig { returns(ActiveSupport::HashWithIndifferentAccess) }
+  def with_indifferent_access; end
+
   sig do
     type_parameters(:K2, :V2, :ConflictRes).params(
       other_hash: T::Hash[T.type_parameter(:K2), T.type_parameter(:V2)],
@@ -152,6 +156,7 @@ class Hash
   end
   def deep_merge!(other_hash, &block); end
 
+  # @version >= 7.2
   sig { params(other: T.untyped).returns(T::Boolean) }
   def deep_merge?(other); end
 end
--- a/rbi/annotations/activesupport.rbi
+++ b/rbi/annotations/activesupport.rbi
@@ -36,7 +36,7 @@ class ActiveSupport::HashWithIndifferentAccess < Hash
       block: T.nilable(
         T.proc.params(
           key: String,
-          old_val: V,
+          old_val: T.untyped,
           new_val: T.type_parameter(:V2),
         ).returns(T.type_parameter(:ConflictRes))
       ),
@@ -50,7 +50,7 @@ class ActiveSupport::HashWithIndifferentAccess < Hash
       block: T.nilable(
         T.proc.params(
           key: String,
-          old_val: V,
+          old_val: T.untyped,
           new_val: T.type_parameter(:V2),
         ).returns(T.type_parameter(:ConflictRes))
       ),
@@ -144,7 +144,7 @@ class Hash
           new_val: T.type_parameter(:V2),
         ).returns(T.type_parameter(:ConflictRes))
       ),
-    ).returns(T::Hash[T.any(K, T.type_parameter(:K2)), T.any(V, T.type_parameter(:ConflictRes))])
+    ).returns(T::Hash[T.any(K, T.type_parameter(:K2)), T.any(V, T.type_parameter(:V2), T.type_parameter(:ConflictRes))])
   end
   def deep_merge(other_hash, &block); end
 
@@ -158,7 +158,7 @@ class Hash
           new_val: T.type_parameter(:V2),
         ).returns(T.type_parameter(:ConflictRes))
       ),
-    ).returns(T::Hash[T.any(K, T.type_parameter(:K2)), T.any(V, T.type_parameter(:ConflictRes))])
+    ).returns(T::Hash[T.any(K, T.type_parameter(:K2)), T.any(V, T.type_parameter(:V2), T.type_parameter(:ConflictRes))])
   end
   def deep_merge!(other_hash, &block); end
 
--- a/rbi/annotations/activesupport.rbi
+++ b/rbi/annotations/activesupport.rbi
@@ -21,8 +21,6 @@ class ActiveSupport::EnvironmentInquirer
 end
 
 class ActiveSupport::HashWithIndifferentAccess < Hash
-  extend T::Generic
-
   K = type_member { {fixed: T.any(String, Symbol)} }
   V = type_member { {fixed: T.untyped} }
   Elem = type_member { {fixed: T.untyped} }
--- a/rbi/annotations/activesupport.rbi
+++ b/rbi/annotations/activesupport.rbi
@@ -21,6 +21,12 @@ class ActiveSupport::EnvironmentInquirer
 end
 
 class ActiveSupport::HashWithIndifferentAccess < Hash
+  extend T::Generic
+
+  K = type_member {{fixed: T.any(String, Symbol)}}
+  V = type_member {{fixed: T.untyped}}
+  Elem = type_member {{fixed: T.untyped}}
+
   sig { returns(ActiveSupport::HashWithIndifferentAccess) }
   def with_indifferent_access; end
 
--- a/rbi/annotations/activesupport.rbi
+++ b/rbi/annotations/activesupport.rbi
@@ -20,6 +20,42 @@ class ActiveSupport::EnvironmentInquirer
   def local?; end
 end
 
+class ActiveSupport::HashWithIndifferentAccess < Hash
+  sig { returns(ActiveSupport::HashWithIndifferentAccess) }
+  def with_indifferent_access; end
+
+  sig do
+    type_parameters(:V2, :ConflictRes).params(
+      other_hash: T::Hash[T.untyped, T.type_parameter(:V2)],
+      block: T.nilable(
+        T.proc.params(
+          key: String,
+          old_val: V,
+          new_val: T.type_parameter(:V2),
+        ).returns(T.type_parameter(:ConflictRes))
+      ),
+    ).returns(ActiveSupport::HashWithIndifferentAccess)
+  end
+  def deep_merge(other_hash, &block); end
+
+  sig do
+    type_parameters(:V2, :ConflictRes).params(
+      other_hash: T::Hash[T.untyped, T.type_parameter(:V2)],
+      block: T.nilable(
+        T.proc.params(
+          key: String,
+          old_val: V,
+          new_val: T.type_parameter(:V2),
+        ).returns(T.type_parameter(:ConflictRes))
+      ),
+    ).returns(ActiveSupport::HashWithIndifferentAccess)
+  end
+  def deep_merge!(other_hash, &block); end
+
+  sig { params(other: T.untyped).returns(T::Boolean) }
+  def deep_merge?(other); end
+end
+
 module ActiveSupport::Testing::SetupAndTeardown::ClassMethods
   sig { params(args: T.untyped, block: T.nilable(T.proc.bind(T.untyped).void)).void }
   def setup(*args, &block); end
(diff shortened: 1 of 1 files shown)


rbi/annotations/activesupport.rbi
```diff
@@ -20,6 +20,42 @@ class ActiveSupport::EnvironmentInquirer
...
+          old_val: V,
+          new_val: T.type_parameter(:V2),
+        ).returns(T.type_parameter(:ConflictRes))
+      ),
+    ).returns(ActiveSupport::HashWithIndifferentAccess)
+  end
+  def deep_merge!(other_hash, &block); end
+
+  sig { params(other: T.untyped).returns(T::Boolean) }
+  def deep_merge?(other); end
```
KaanOzkan, line 57: I think this is from v7.1 onwards, we could annotate it with `#@version  >= 7.2` https://github.com/rails/rails/commit/43b980368a7628fac95cc4f673e0dfbcee77c10b

rbi/annotations/activesupport.rbi
```diff
@@ -20,6 +20,42 @@ class ActiveSupport::EnvironmentInquirer
   def local?; end
 end
 
+class ActiveSupport::HashWithIndifferentAccess < Hash
+  sig { returns(ActiveSupport::HashWithIndifferentAccess) }
+  def with_indifferent_access; end
```
KaanOzkan, line 25: The signature is being added to `HashWithIndifferentAccess` but you mentioned wanting to use this with `Hash.with_indifferent_access`, I'm not sure this signature will be picked up in that case, can you verify please?

rbi/annotations/activesupport.rbi
```diff
@@ -20,6 +20,42 @@ class ActiveSupport::EnvironmentInquirer
...
+
+  sig do
+    type_parameters(:V2, :ConflictRes).params(
+      other_hash: T::Hash[T.untyped, T.type_parameter(:V2)],
+      block: T.nilable(
+        T.proc.params(
+          key: String,
+          old_val: V,
```
KaanOzkan, line 33: This value, `V`, is undefined. Please locally verify these annotations work before upstreaming.

pyrkhanton:
"I have signed the CLA!"

rbi/annotations/activesupport.rbi
pyrkhanton: `V` should be available from [sorbet-typed](https://github.com/sorbet/sorbet-typed/blob/master/lib/activesupport/all/activesupport.rbi#L1783)

Even though it's fixed to `T.untyped` does it make sense to have it!?

Btw I'm testing the signatures on a project by adding them into `sorbet/rbi/shims/gems/activesupport.rbi` and here I don't have problems with visibility o `V` type. Pls share If I could do this in a better way.

rbi/annotations/activesupport.rbi
pyrkhanton: This one I think also makes sense to keep for chaining calls like:
```ruby
hwia = {a: 'b'}.with_indifferent_access
hwia_on_hwia = hwia.with_indifferent_access
```

But yeap, looks like I forgot to add this on `Hash` within this PR, I'll add this.
If not add the method sig on `ActiveSupport::HashWithIndifferentAccess` second like from example would result in `T.untyped`

rbi/annotations/activesupport.rbi
pyrkhanton: Nice catch. I'll add version.

rbi/annotations/activesupport.rbi
pyrkhanton: PS used example to check signatures
```ruby
# typed: strict
# frozen_string_literal: true

class TestSorbet
  extend T::Sig

  sig { params(h: T::Hash[Symbol, Integer]).void }
  def test_meth_hash(h)
    plain = T.let({ a: "a" }, T::Hash[Symbol, String])
    hwia = { a: "a" }.with_indifferent_access

    plain_res = h.deep_merge(plain)
    pp plain_res
    hwia_res = h.deep_merge(hwia)
    pp hwia_res

    conflict_plain_res = h.deep_merge(plain) do |k, o, n|
      abc = ("a".."z").to_a.each_with_index.to_a.to_h
      new_val = T.must(abc[k.to_s]) * 99.99
      new_val
    end
    pp conflict_plain_res

    conflict_hwia_res = h.deep_merge(hwia) do |k, o, n|
      abc = ("a".."z").to_a.each_with_index.to_a.to_h
      new_val = T.must(abc[k.to_s]) * 99.99
      new_val
    end
    pp conflict_hwia_res
  end

  sig { params(h: ActiveSupport::HashWithIndifferentAccess).void }
  def test_meth_hwia(h)
    plain = T.let({ a: "a" }, T::Hash[Symbol, String])
    { b: "b" }.deep_merge?(plain)
    hwia = { a: "a" }.with_indifferent_access
    hwia.deep_merge?(h)

    hwia_on_hwia = hwia.with_indifferent_access

    plain_res = h.deep_merge(plain)
    pp plain_res
    hwia_res = h.deep_merge(hwia)
    pp hwia_res

    conflict_plain_res = h.deep_merge(plain) do |k, o, n|
      abc = ("a".."z").to_a.each_with_index.to_a.to_h
      new_val = T.must(abc[k.to_s]) * 99.99
      new_val
    end
    pp conflict_plain_res

    conflict_hwia_res = h.deep_merge(hwia) do |k, o, n|
      abc = ("a".."z").to_a.each_with_index.to_a.to_h
      new_val = T.must(abc[k.to_s]) * 99.99
      new_val
    end
    pp conflict_hwia_res
  end
end

```

KaanOzkan: > V should be available from [sorbet-typed](https://github.com/sorbet/sorbet-typed/blob/master/lib/activesupport/all/activesupport.rbi#L1783)

Ah that makes sense. RBI Central is used by Tapioca to pull annotations when you run `tapioca annotations` so it has no relation to sorbet-typed and we can't expect users to have it installed. So you need to test the annotation on its own, without any definitions from `sorbet-typed`. In this particular case I think you can just say `T.untyped` instead of using a generic variable.

pyrkhanton: Now I see. Those `sorbet-typed` definitions I saw was some legacy definitions.
It would make sense to have all just here.

Does it make sense to follow `Hash` generic definition like it's done in [sorbet](https://github.com/sorbet/sorbet/blob/master/rbi/core/hash.rbi#L148-L150)
To have comparable signatures for `Hash` and `HashWithIndifferentAccess` even though for last one we'd have them `fixed` in place.
Basically I'm thinking that it's a good idea to migrate [definition](https://github.com/sorbet/sorbet-typed/blob/master/lib/activesupport/all/activesupport.rbi#L1782-L1784) here and be able to use `K` and `V` for methods signatures.

KaanOzkan: Looks like they are just `fixed`, I'd prefer having the type directly in the signature instead of using `K`, `V`, it'll be easier for someone to see the type, i.e. on hover using the LSP.

rbi/annotations/activesupport.rbi
```diff
@@ -87,6 +130,41 @@ class Hash
...
+    type_parameters(:K2, :V2, :ConflictRes).params(
+      other_hash: T::Hash[T.type_parameter(:K2), T.type_parameter(:V2)],
+      block: T.nilable(
+        T.proc.params(
+          key: T.any(K, T.type_parameter(:K2)),
+          old_val: V,
+          new_val: T.type_parameter(:V2),
+        ).returns(T.type_parameter(:ConflictRes))
+      ),
+    ).returns(T::Hash[T.any(K, T.type_parameter(:K2)), T.any(V, T.type_parameter(:ConflictRes))])
```
KaanOzkan:
```suggestion
    ).returns(T::Hash[T.any(K, T.type_parameter(:K2)), T.any(V, T.type_parameter(:V2), T.type_parameter(:ConflictRes))])
```

rbi/annotations/activesupport.rbi
```diff
@@ -87,6 +130,41 @@ class Hash
...
```
KaanOzkan:
```suggestion
    ).returns(T::Hash[T.any(K, T.type_parameter(:K2)), T.any(V, T.type_parameter(:V2), T.type_parameter(:ConflictRes))])
```

rbi/annotations/activesupport.rbi
```diff
@@ -20,6 +20,49 @@ class ActiveSupport::EnvironmentInquirer
...
+  end
+  def deep_merge(other_hash, &block); end
+
+  sig do
+    type_parameters(:V2, :ConflictRes).params(
+      other_hash: T::Hash[T.untyped, T.type_parameter(:V2)],
+      block: T.nilable(
+        T.proc.params(
+          key: String,
+          old_val: V,
```
KaanOzkan:
```suggestion
          old_val: T.untyped,
```

rbi/annotations/activesupport.rbi
pyrkhanton: FYI when I hover for LSP signatures when having `old_val: V` it's computed for me (see `arg1: T.untyped`)
Using VSCode

Image

I thought it would be better for understanding to have kind of reference that `old_val` is a type of current object values `V` even thought it would be always `T.untyped`.
But I'm ok to change it explicitly to `T.untyped`, let me do this.

rbi/annotations/activesupport.rbi
pyrkhanton: Applied the change.

rbi/annotations/activesupport.rbi
pyrkhanton: Applied the change.

rbi/annotations/activesupport.rbi
pyrkhanton: Applied the change.

rbi/annotations/activesupport.rbi
pyrkhanton: Changed to explicit `old_val: T.untyped`.

rbi/annotations/activesupport.rbi
KaanOzkan: Do you still want to add this?

KaanOzkan: Oh I didn't realize the next diff. Nvm.

rbi/annotations/activesupport.rbi
```diff
@@ -20,6 +20,49 @@ class ActiveSupport::EnvironmentInquirer
   def local?; end
 end
 
+class ActiveSupport::HashWithIndifferentAccess < Hash
+  extend T::Generic
+
+  K = type_member {{fixed: T.any(String, Symbol)}}
+  V = type_member {{fixed: T.untyped}}
```
KaanOzkan: Do we still need this?

rbi/annotations/activesupport.rbi
```diff
@@ -87,6 +130,41 @@ class Hash
...
+  sig { returns(ActiveSupport::HashWithIndifferentAccess) }
+  def with_indifferent_access; end
+
+  sig do
+    type_parameters(:K2, :V2, :ConflictRes).params(
+      other_hash: T::Hash[T.type_parameter(:K2), T.type_parameter(:V2)],
+      block: T.nilable(
+        T.proc.params(
+          key: T.any(K, T.type_parameter(:K2)),
+          old_val: V,
```
KaanOzkan, line 141: I don't think `K` and `V` are defined. If it's going to be fixed to `T.untyped` we can just inline it. Similarly for `deep_merge!`

rbi/annotations/activesupport.rbi
```diff
@@ -20,6 +20,49 @@ class ActiveSupport::EnvironmentInquirer
   def local?; end
 end
 
+class ActiveSupport::HashWithIndifferentAccess < Hash
+  extend T::Generic
+
+  K = type_member {{fixed: T.any(String, Symbol)}}
```
KaanOzkan: Formatting issues https://github.com/Shopify/rbi-central/actions/runs/24887364599/job/73208971561?pr=388#step:4:28

rbi/annotations/activesupport.rbi
pyrkhanton: For `Hash` we have generic definition in Sorbet [core types](https://github.com/sorbet/sorbet/blob/master/rbi/core/hash.rbi#L147-L150).
Hope we can rely on this.
If that's true `K` and `V` should be available and we can use it.

rbi/annotations/activesupport.rbi
pyrkhanton: Given that for `Hash` we have generic definition in Sorbet [core types](https://github.com/sorbet/sorbet/blob/master/rbi/core/hash.rbi#L147-L150).
And `ActiveSupport::HashWithIndifferentAccess` inherited from `Hash` once we define fixed types they would be evaluated on child methods.
For example [#values](https://github.com/sorbet/sorbet/blob/master/rbi/core/hash.rbi#L1520-L1521) would return `T::Array[V]` -> `T::Array[T.untyped]`.
Same should be for other methods using `K` also.

rbi/annotations/activesupport.rbi
```diff
@@ -20,6 +20,49 @@ class ActiveSupport::EnvironmentInquirer
   def local?; end
 end
 
+class ActiveSupport::HashWithIndifferentAccess < Hash
+  extend T::Generic
```
KaanOzkan:
```suggestion
```

Fails at runtime and isn't necessary

KaanOzkan (approved):

KaanOzkan:
Thanks @pyrkhanton appreciate you sticking with me during the review back and forth.

PR #392, opened 2026-04-28
Bump tapioca to 0.19.1 and sorbet to 0.6.13164

soutaro:
## Summary

- `tapioca 0.18.0` calls `has_rest` on `T::Private::Methods::Signature`, which is no longer available in newer `sorbet-runtime` releases. Test fixtures that run `bundle install` resolve against the latest `sorbet-runtime` (currently `0.6.13164`) and crash with `NoMethodError`, breaking `CheckStaticTest`, `RepoTest`, and `CheckTest` on CI (e.g. https://github.com/Shopify/rbi-central/actions/runs/25031509086/job/73313926311).
- Bump `tapioca` to `0.19.1` (compatible with newer `sorbet-runtime`) and raise the gemspec floor to `>= 0.19.1` so consumers can't resolve back to `0.18.0`.
- Bump the `sorbet*` family from `0.6.13071` to `0.6.13164` so our lockfile matches what fixtures install at runtime.

`tapioca 0.19.x` introduces a new transitive dependency on `rubydex (>= 0.1.0.beta10)` (currently a prerelease). Bundler resolves this without `--pre` because no stable version satisfies the constraint.

## Test plan

- [ ] CI green (the previously failing `CheckStaticTest`, `RepoTest`, and `CheckTest` should pass)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -10,7 +10,7 @@ PATH
       rubocop-sorbet (>= 0.6.11)
       sorbet-static-and-runtime (>= 0.5.11922)
       spoom (>= 1.2.0)
-      tapioca (>= 0.17.7)
+      tapioca (>= 0.19.1)
       thor (>= 1.2.1)
 
 GEM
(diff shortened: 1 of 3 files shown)


soutaro:
Folded into #391 since the IMAP fix can't be tested in CI without this dependency bump.

PR #393, opened 2026-04-30
Add Rails.event sig

bdewater-thatch:
### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem name: Railties
* Gem version: 8.1
* Upstream: https://github.com/rails/rails/commit/071670b2697a21f3eae629a679f9b45b02ac1072

### DIFF
--- a/rbi/annotations/railties.rbi
+++ b/rbi/annotations/railties.rbi
@@ -17,6 +17,10 @@ module Rails
     sig { returns(ActiveSupport::ErrorReporter) }
     def error; end
 
+    # @version >= 8.1.0.beta1
+    sig { returns(ActiveSupport::EventReporter) }
+    def event; end
+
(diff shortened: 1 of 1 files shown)


rbi/annotations/railties.rbi
```diff
@@ -14,6 +14,10 @@ module Rails
     sig { returns(ActiveSupport::EnvironmentInquirer) }
     def env; end
 
+    # @version >= 8.1.0.beta1
+    sig { returns(ActiveSupport::EventReporter) }
+    def event; end
```
KaanOzkan, line 22: Nit, Could you move this after `error`, it'll be in alphabetical order. Thanks.

KaanOzkan (approved):

PR #395, opened 2026-05-08
Add Mock class for Mocha

thecharlesjenkins:
Mock was being referred to without definition causing an error with sorbet:

sorbet/rbi/annotations/mocha.rbi:8: Unable to resolve constant Mock https://srb.help/5002
     8 |  sig { params(arguments: T.untyped).returns(Mocha::Mock) }

Add the definition of the Mock class to resolve this error as well as make the type more complete.

### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem name: Mocha
* Gem version: 3.1.0
* Gem source: https://github.com/freerange/mocha/blob/main/lib/mocha/mock.rb
* Gem API doc: https://mocha.jamesmead.org/Mocha/Mock.html
* Tapioca version: 0.18.0
* Sorbet version: 0.6.13096

### DIFF
--- a/rbi/annotations/mocha.rbi
+++ b/rbi/annotations/mocha.rbi
@@ -1,5 +1,32 @@
 # typed: true
 
+class Mocha::Mock
+  sig do
+    params(
+      method_name_or_hash: T.any(Symbol, String, T::Hash[Symbol, T.untyped]),
+      backtrace: T.nilable(T::Array[String])
+    ).returns(Mocha::Expectation)
+  end
+  def expects(method_name_or_hash, backtrace = nil); end
+
+  sig do
+    params(
+      method_name_or_hash: T.any(Symbol, String, T::Hash[Symbol, T.untyped]),
+      backtrace: T.nilable(T::Array[String])
+    ).returns(Mocha::Expectation)
+  end
+  def stubs(method_name_or_hash, backtrace = nil); end
+
+  sig { params(method_names: Symbol).void }
+  def unstub(*method_names); end
+
+  sig { params(responder: Object).returns(T.self_type) }
+  def responds_like(responder); end
+
+  sig { params(responder_class: T::Class[T.untyped]).returns(T.self_type) }
+  def responds_like_instance_of(responder_class); end
+end
+
 module Mocha::API
   sig { params(arguments: T.untyped).returns(Mocha::Mock) }
   def mock(*arguments); end


rbi/annotations/mocha.rbi
```diff
@@ -1,5 +1,38 @@
 # typed: true
 
+class Mocha::Mock
```
KaanOzkan, line 3: > Mock was being referred to without definition causing an error with sorbet

`Mocha::Mock` should've been discovered by `tapioca gem` and be part of the gem RBIs for `mocha`. If it's still not visible after running `tapioca gem mocha` you might need a `require` to load it for Tapioca to see, https://github.com/Shopify/tapioca#manually-requiring-parts-of-a-gem.

rbi/annotations/mocha.rbi
```diff
@@ -1,5 +1,38 @@
...
+    sig do
```
KaanOzkan: The indentation in this class is off, it needs to be 2 spaces. You can run `bundle exec repo check rubocop` locally to check.

rbi/annotations/mocha.rbi
```diff
@@ -1,5 +1,38 @@
...
+        method_name_or_hash: T.any(Symbol, String, T::Hash[Symbol, T.untyped]),
+        backtrace: T.nilable(T::Array[String])
+      ).returns(Mocha::Expectation)
+    end
+    def stubs(method_name_or_hash, backtrace = nil); end
+
+    sig { params(method_names: Symbol).void }
+    def unstub(*method_names); end
+
+    sig { params(responder: T.all(Object, T.proc.params(arg0: Symbol).returns(T::Boolean))).returns(T.self_type) }
```
KaanOzkan:
```suggestion
    sig { params(responder: Object).returns(T::Boolean))).returns(T.self_type) }
```

I don't think we need to intersect it with `Object` since `Object` covers it. Looks like this accepts anything with `respond_to?` so `Object` makes sense.

rbi/annotations/mocha.rbi
```diff
@@ -1,5 +1,38 @@
...
+    def responds_like(responder); end
+
+    sig { params(responder_class: T::Class[T.untyped]).returns(T.self_type) }
+    def responds_like_instance_of(responder_class); end
+
+    sig { params(mockery: T.untyped, assertion_counter: T.untyped, name: T.untyped, receiver: T.untyped).void }
+    def initialize(mockery, assertion_counter, name = nil, receiver = nil); end
```
KaanOzkan:
```suggestion
```

This signature doesn't seem to be bringing any value, also method is marked `@private` so it's not part of the advertised public API. If you do need this resolving the `tapioca gem mocha` issue is better.

rbi/annotations/mocha.rbi
```diff
@@ -1,5 +1,38 @@
...
+
+    alias_method :quacks_like, :responds_like
+    alias_method :quacks_like_instance_of, :responds_like_instance_of
```
KaanOzkan:
```suggestion
```

Similarly these are already generated by `tapioca gem`, redefining them like this is a [type checking error](https://github.com/Shopify/rbi-central/actions/runs/25537483569/job/75034054153?pr=395#step:4:91).

KaanOzkan (commented):
Thanks, the types will be useful.

rbi/annotations/mocha.rbi
thecharlesjenkins: Hmm okay thank you! It wasn't being automatically discovered

rbi/annotations/mocha.rbi
```diff
@@ -1,5 +1,32 @@
...
+      method_name_or_hash: T.any(Symbol, String, T::Hash[Symbol, T.untyped]),
+      backtrace: T.nilable(T::Array[String])
+    ).returns(Mocha::Expectation)
+  end
+  def stubs(method_name_or_hash, backtrace = nil); end
+
+  sig { params(method_names: Symbol).void }
+  def unstub(*method_names); end
+
+  sig { params(responder: Object).returns(T::Boolean).returns(T.self_type) }
```
KaanOzkan:
```suggestion
  sig { params(responder: Object).returns(T.self_type) }
```

Could you ensure all the sigs work in your project? It'd be a good manual test for the underlying types.

thecharlesjenkins: Yes, all of the sigs do work in my project

KaanOzkan (approved):

PR #396, opened 2026-05-08
Add Rails::Railtie.rake_tasks sig

frewsxcv:
## Summary

Adds a sig for `Rails::Railtie.rake_tasks` that binds the block to `Rake::DSL` and types the yielded `app` argument as `Rails::Application`.

In Rails, [`run_tasks_blocks`](https://github.com/rails/rails/blob/main/railties/lib/rails/railtie.rb) executes registered blocks via `instance_exec(app, &block)` after `extend Rake::DSL`, so `self` inside the block legitimately responds to `task`, `namespace`, `desc`, etc.

Without this annotation, idiomatic code like:

```ruby
class MyEngine < ::Rails::Engine
  rake_tasks do |_app|
    task 'db:schema:dump': 'strong_migrations:alphabetize_columns'
  end
end
```

fails type checking under `# typed: true` with:

```
Method `task` does not exist on `T.class_of(MyEngine)` https://srb.help/7003
```

forcing users to fall back to `T.unsafe(self).task(...)` (which disables checking entirely) or to drop the file to `# typed: false`.

This mirrors the existing DSL-binding pattern already used in this file for `Rails::Engine#routes` (binds to `ActionDispatch::Routing::Mapper`), `Rails::Application#configure` (binds to `Rails::Application`), and `Rails::Railtie::Configuration#to_prepare` (binds to `ActiveSupport::Reloader`).

`rake` is already listed in the `requires` for railties in `index.json`, so `Rake::DSL` is available without index changes.

## Test plan

- [x] `bundle exec repo check static` passes
- [x] `bundle exec repo check runtime` passes
- [x] `bundle exec repo check rubocop` passes

### DIFF
--- a/rbi/annotations/railties.rbi
+++ b/rbi/annotations/railties.rbi
@@ -64,6 +64,11 @@ end
 class Rails::Railtie
   sig { params(block: T.proc.bind(Rails::Railtie).void).void }
   def configure(&block); end
+
+  class << self
+    sig { params(block: T.proc.bind(Rake::DSL).params(app: Rails::Application).void).void }
+    def rake_tasks(&block); end
+  end
 end
 
 class Rails::Railtie::Configuration


KaanOzkan (approved):

KaanOzkan:
@frewsxcv thanks for the PR, could you drop the empty commit?

frewsxcv:
@KaanOzkan You should be able to squash the pull request in GitHub, which would remove the empty commit from the `main` commit timeline.

KaanOzkan:
@frewsxcv that option is disabled on this repo since we want to keep the history, although in this case it'd be good to have. Can you also rebase the branch on `main`? There has been new changes and I'll re-trigger the CI afterwards.

EDIT: I ended up rebasing it

frewsxcv:
Screenshot 2026-05-20 at 10 31 53 AM

KaanOzkan:
I'm aware that you can do it through GitHub UI but as I said it's disabled for this repo.

PR #397, opened 2026-05-13
Add sigs for Active Record becomes methods

bdewater-thatch:
### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem name: Active Record
* Gem source: https://github.com/rails/rails/blob/5de2d28389a50a2e0d669f2d157eb6f89b59103a/activerecord/lib/active_record/persistence.rb#L487


KaanOzkan (approved):

KaanOzkan:
Looks like it needs a rebase.

PR #399, opened 2026-05-14
Add ActiveModel/Record attribute methods sigs

bdewater-thatch:
### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other: 

### Changes

I noticed people writing code assuming these methods return symbols instead of strings, so it wasn't giving the results they were expecting. These signatures will now flag that kind of thing, e.g. `app/models/member.rb:139: Expected String but found Symbol(:address_line1) for argument arg0 https://srb.help/7002`

* Gem name: Active Model / Record
* Gem API doc:
  * https://api.rubyonrails.org/classes/ActiveModel/Attributes.html
  * https://api.rubyonrails.org/classes/ActiveModel/Dirty.html
  * https://api.rubyonrails.org/classes/ActiveRecord/AttributeMethods/ClassMethods.html
  * https://api.rubyonrails.org/classes/ActiveRecord/AttributeMethods/Dirty.html

### DIFF
--- a/rbi/annotations/activemodel.rbi
+++ b/rbi/annotations/activemodel.rbi
@@ -1,5 +1,35 @@
 # typed: true
 
+module ActiveModel::Attributes
+  sig { returns(T::Array[String]) }
+  def attribute_names; end
+
+  sig { returns(T::Hash[String, T.untyped]) }
+  def attributes; end
+end
+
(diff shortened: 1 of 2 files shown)


KaanOzkan (approved):
This needs a rebase too 😬

PR #401, opened 2026-05-15
Fix couple of AM/AR Dirty methods returning HWIA instead of Hash

bdewater-thatch:
### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other: 

### Changes

Follow-up to https://github.com/Shopify/rbi-central/pull/399 - turns out some of these return HWIA instead of a string-keyed Hash 🤦


KaanOzkan (approved):

KaanOzkan:
Can't edit the PR, can you rebase and let me know?

bdewater-thatch:
@KaanOzkan done!

PR #408, opened 2026-07-06
Add ActiveRecord query assertions RBI

KaanOzkan:
### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem name: activerecord
* Gem source: https://github.com/rails/rails/blob/main/activerecord/lib/active_record/testing/query_assertions.rb#L24

### DIFF
--- a/rbi/annotations/activerecord.rbi
+++ b/rbi/annotations/activerecord.rbi
@@ -37,6 +37,53 @@ class ActiveRecord::Migration
   include ActiveRecord::ConnectionAdapters::DatabaseStatements
 end
 
+# @version >= 7.2.0
+module ActiveRecord::Assertions::QueryAssertions
+  sig do
+    type_parameters(:R)
+      .params(
+        count: T.nilable(Integer),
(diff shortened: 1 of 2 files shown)


rbi/annotations/activerecord.rbi
```diff
@@ -37,6 +37,53 @@ class ActiveRecord::Migration
   include ActiveRecord::ConnectionAdapters::DatabaseStatements
 end
 
+# @version >= 7.2.0
```
Morriar, line 40: Could we include the 7.2 prereleases here?

```suggestion
# @version >= 7.2.0.beta1
```

This file exists in Rails v7.2.0.beta1/beta2/beta3/rc1 too, and `Gem::Version` sorts those below `7.2.0`. With the current range, `tapioca annotations` would strip these methods for projects pinned to a 7.2 prerelease even though the runtime API is present.

jesse-shopify (approved):

PR #409, opened 2026-07-07
Include ActiveRecord query assertions in test cases

KaanOzkan:
Makes the annotations added in https://github.com/Shopify/rbi-central/pull/408 more useful

### Type of Change

- [ ] Add RBI for a new gem
- [ ] Other: 

### Changes

* Gem source: Defined in railties but I think this annotation is the best place to put it https://github.com/rails/rails/blob/main/railties/lib/rails/test_help.rb#L15-L20

### DIFF
--- a/rbi/annotations/activerecord.rbi
+++ b/rbi/annotations/activerecord.rbi
@@ -84,6 +84,12 @@ module ActiveRecord::Assertions::QueryAssertions
   def assert_no_queries_match(match, include_schema: false, &block); end
 end
 
+# @version >= 7.2.0
+class ActiveSupport::TestCase
+  # @shim: Rails includes query assertions into ActiveSupport::TestCase from rails/test_help when ActiveRecord is loaded.
+  include ActiveRecord::Assertions::QueryAssertions
+end
+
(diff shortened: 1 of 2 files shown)


rbi/annotations/activerecord.rbi
```diff
@@ -84,6 +84,11 @@ module ActiveRecord::Assertions::QueryAssertions
   def assert_no_queries_match(match, include_schema: false, &block); end
 end
 
+class ActiveSupport::TestCase
```
Morriar, line 88: This needs the same version gate as `ActiveRecord::Assertions::QueryAssertions`. The module above is only present on ActiveRecord >= 7.2.0, but this `ActiveSupport::TestCase` include would still be pulled for older ActiveRecord versions after version filtering, leaving an `include ActiveRecord::Assertions::QueryAssertions` reference to a constant that was filtered out.

```suggestion
# @version >= 7.2.0
class ActiveSupport::TestCase
```

KaanOzkan: Done.

Morriar (approved):

PR #417, opened 2026-09-01
Add dev.yml and upgrade Ruby to v4.0.6

vinistock:
Add `dev.yml` so we can manage dependencies with Nix and upgrade to the latest Ruby.


KaanOzkan (approved):

.ruby-version
```diff
@@ -1 +1 @@
-3.4.3
+4.0.6
```
Morriar: Let's also update `gem/.ruby-version` to `4.0.6` so the gem workspace doesn't keep using the old Ruby. The README still tells contributors to work from `gem/`, so commands run there will pick up `3.4.3` while the root and CI use `4.0.6`.

jesse-shopify (approved):

PR #419, opened 2026-09-02
Add annotations for i18n

mokevnin:
### Type of Change

- [x] Add RBI for a new gem

### Changes

* Gem name: i18n
* Gem version: 1.15.2
* Gem source: https://github.com/ruby-i18n/i18n/blob/v1.15.2/lib/i18n.rb#L96-L108 (the `module_eval` delegator block) and https://github.com/ruby-i18n/i18n/blob/v1.15.2/lib/i18n/config.rb
* Gem API doc: https://rubydoc.info/gems/i18n/1.15.2/I18n/Base
* Tapioca version: 0.19.1
* Sorbet version: 0.6.13164

`I18n::Base` defines `locale`, `default_locale`, `available_locales`,
`default_separator` and `enforce_available_locales` (plus their writers) with
`module_eval` over a hardcoded list, each forwarding to `I18n::Config`. Tapioca
emits them without signatures, so `I18n.locale` and friends are `T.untyped`.
Only methods whose types are fixed by the gem's own source are annotated:

* `config` / `writable_config` → `I18n::Config` — both return the memoized `I18n::Config` (or its dup).
* `locale` → `Symbol` — `Config#locale` falls back to `default_locale`, which is `@@default_locale ||= :en`.
* `default_locale` → `Symbol` — same `||= :en` fallback, so it is never nil.
* `locale=` / `default_locale=` → `T.nilable(T.any(String, Symbol))` — the writers do `locale && locale.to_sym`.
* `available_locales` → `T::Array[Symbol]` — the writer stores `Array(locales).map(&:to_sym)`, and `Backend::Simple#available_locales` returns the symbol keys of the translations hash.
* `available_locales=` → accepts one locale, an array of locales, or nil, since it applies `Array()` then `to_sym`.
* `default_separator` → `String` — `@@default_separator ||= '.'`.
* `enforce_available_locales` → `T::Boolean` — class variable initialized to `true`.

Deliberately left out because their types are genuinely open in the gem:

* `backend` / `backend=` — any object implementing the backend protocol; `I18n::Backend::Base` is not enforced.
* `exception_handler` / `exception_handler=` — documented as anything responding to `#call`.
* `load_path` / `load_path=` — the gem states custom implementations only need to "behave like a Ruby Array", so `T::Array` would over-promise.
* `translate` and friends — return type depends on the stored translation.

`bundle exec repo check --gem --ref upstream/main` passes locally (index, rubocop, rubygems, runtime, static).

### DIFF
--- a/rbi/annotations/i18n.rbi
+++ b/rbi/annotations/i18n.rbi
@@ -1,0 +1,42 @@
+# typed: true
+
+module I18n::Base
+  sig { returns(T::Array[Symbol]) }
+  def available_locales; end
+
+  sig do
+    params(value: T.nilable(T.any(String, Symbol, T::Array[T.any(String, Symbol)])))
+      .returns(T.nilable(T.any(String, Symbol, T::Array[T.any(String, Symbol)])))
+  end
+  def available_locales=(value); end
+
+  sig { returns(I18n::Config) }
+  def config; end
+
+  sig { returns(Symbol) }
+  def default_locale; end
+
+  sig { params(value: T.nilable(T.any(String, Symbol))).returns(T.nilable(T.any(String, Symbol))) }
+  def default_locale=(value); end
+
+  sig { returns(String) }
+  def default_separator; end
+
+  sig { params(value: String).returns(String) }
+  def default_separator=(value); end
+
+  sig { returns(T::Boolean) }
+  def enforce_available_locales; end
+
+  sig { params(value: T::Boolean).returns(T::Boolean) }
+  def enforce_available_locales=(value); end
+
+  sig { returns(Symbol) }
+  def locale; end
+
+  sig { params(value: T.nilable(T.any(String, Symbol))).returns(T.nilable(T.any(String, Symbol))) }
+  def locale=(value); end
+
+  sig { returns(I18n::Config) }
+  def writable_config; end
+end
--- a/index.json
+++ b/index.json
@@ -74,6 +74,7 @@
     ]
   },
   "graphql": {},
+  "i18n": {},
   "kredis": {
     "dependencies": [
       "activemodel",


rbi/annotations/i18n.rbi
```diff
@@ -0,0 +1,42 @@
...
+  def enforce_available_locales=(value); end
+
+  sig { returns(Symbol) }
+  def locale; end
+
+  sig { params(value: T.nilable(T.any(String, Symbol))).returns(T.nilable(T.any(String, Symbol))) }
+  def locale=(value); end
+
+  sig { returns(I18n::Config) }
+  def writable_config; end
```
Morriar: I think we need version guards here. `I18n::Base#writable_config` only exists starting in i18n `1.15.0`, but this annotation will be pulled for older i18n versions too. For example, `1.14.8` has `I18n::Base` but does not define `writable_config`, so this would add a runtime-missing method for projects pinned below `1.15.0`.

Could we guard this method? I also wonder if we should guard the whole `I18n::Base` block with `# @version >= 0.8.0`, since older i18n releases do not define that module.

```ruby
# @version >= 1.15.0
sig { returns(I18n::Config) }
def writable_config; end
```

A regression check would be to run the runtime checker against i18n `1.14.8` and confirm `writable_config` is filtered out.

rbi/annotations/i18n.rbi
```diff
@@ -0,0 +1,42 @@
+# typed: true
+
+module I18n::Base
+  sig { returns(T::Array[Symbol]) }
```
KaanOzkan, line 4: I think this should follow the setter, `T::Array[T.any(String, Symbol)]`, which is also the case in https://github.com/ruby/gem_rbs_collection/blob/31250b67b9dadaeaccdb65c55d64a20dca7a812e/gems/i18n/1.10/i18n.rbs#L19

mokevnin: The setter does accept strings, but it symbolizes on the way in — `@@available_locales = Array(locales).map { |locale| locale.to_sym }` ([config.rb:92](https://github.com/ruby-i18n/i18n/blob/master/lib/i18n/config.rb#L91-L95)) — so nothing a caller sets can come back out as a String.

That is the same shape as `locale=` and `default_locale=`: both take `String | Symbol` and store `locale && locale.to_sym`, and their getters are annotated `Symbol` in this file. Following the setter for `available_locales` alone would break that symmetry, and following it consistently would make `I18n.locale` a union too.

There is a real reason to loosen this one, just a different one: with `@@available_locales` unset the call falls through to `backend.available_locales`, and `backend` is a pluggable protocol. Every backend shipped in the gem returns symbols — `Simple` (translation hash keys), `KeyValue` (`locales.map! { |k| k.to_sym }`), `LazyLoadable` (`locale.to_sym`), `Memoize`/`Chain` delegate — so the union would only be covering third-party backends.

Happy to widen it to `T::Array[T.any(String, Symbol)]` if you would rather the annotation cover those — say the word and I will push it.

KaanOzkan: You’re right about the setter. The backend path isn’t normalized, though:

    backend = Object.new
    def backend.available_locales = ["en"]

    I18n.backend = backend
    I18n.available_locales # => ["en"]

So I think we should keep `T::Array[T.any(String, Symbol)]` for custom backends.

mokevnin: Agreed — the backend path is the real reason, and it settles it. `available_locales` falls back to `backend.available_locales` whenever nothing was set explicitly ([config.rb:77-80](https://github.com/ruby-i18n/i18n/blob/master/lib/i18n/config.rb#L77-L80)), and that list comes back exactly as the backend built it. `available_locales_set` normalizing both ways right below is the gem itself saying the same thing.

Widened to `T::Array[T.any(String, Symbol)]` with a comment recording why, pushed as `053893d`. Also rebased on `main` so the branch is no longer behind.

KaanOzkan (approved):

mokevnin:
Thanks for the thorough review — the backend-path catch was a good one.

PR #422, opened 2026-09-02
Add annotations for the alba gem

mokevnin:
### Type of Change

- [x] Add RBI for a new gem

### Changes

* Gem name: alba
* Gem version: 4.0.0
* Gem source: https://github.com/okuramasafumi/alba/blob/main/lib/alba/resource.rb
* Gem API doc: https://rubydoc.info/gems/alba/Alba/Resource/ClassMethods
* Tapioca version: 0.19.1
* Sorbet version: 0.6.13164

Alba ships no signatures, so tapioca emits every `Alba::Resource::ClassMethods` DSL macro sig-less. This adds signatures for the class-level macros used to define a resource. Every parameter list matches the one tapioca generates; every type is taken from the method body and its YARD doc.

* `attribute` / `attributes` — `name` is `String`/`Symbol`, `if:` is `Symbol` or `Proc` (`Alba::ConditionalAttribute#condition_passes?` branches on exactly those two), extra keywords are the untyped `name => type` pairs handed to `Alba::TypedAttribute`.
* `association` and its aliases `one`, `many`, `has_one`, `has_many` — types taken from the YARD `@param` tags on `association`, which the aliases share; `resource:`/`serializer:` accept a resource class, a `String`, or a `Proc`.
* `nested_attribute` / `nested`, `trait` — block is `class_eval`'d, so it is bound to an untyped self and returns nothing useful.
* `on_error` — `handler` is one of `:raise`, `:ignore`, `:nullify`; the block is called as `on_error.call(error, obj, key, attribute, klass)` and its `[key, value]` return is destructured, hence `T::Array[T.untyped]`.
* `on_nil` — called as `instance_exec(obj, key, attribute, &nil_handler)`.
* `meta`, `layout`, `helper`, `collection_key`, `root_key`, `root_key!`, `root_key_for_collection`, `transform_keys`, `prefer_object_method!`, `prefer_resource_method!` — direct reads of the assignments in each body.

Two notes:

* The block on `attribute` (and on `trait` and `nested_attribute`) is declared non-nilable. It is syntactically optional, but all three raise `ArgumentError` immediately when it is absent, so a call without a block can only ever fail at runtime.
* `transform_keys!` is left unannotated: it returns the anonymous `dup` of the resource class, is marked experimental in the source, and I could not type it honestly.

Instance methods (`serialize`, `as_json`, `serializable_hash`) are deliberately left out of this first pass — `serializable_hash` returns a `Hash` or an `Array` depending on whether the object is a collection, and a union return there would tighten code that type-checks today.

`bundle exec repo check --gem --ref upstream/main` passes locally on all five checks (index, rubocop, rubygems, runtime, static).

### DIFF
--- a/rbi/annotations/alba.rbi
+++ b/rbi/annotations/alba.rbi
@@ -1,0 +1,173 @@
+# typed: true
+
+module Alba::Resource::ClassMethods
+  sig do
+    params(
+      name: T.any(String, Symbol),
+      condition: T.nilable(Proc),
+      resource: T.nilable(T.any(Module, String, Proc)),
+      serializer: T.nilable(T.any(Module, String, Proc)),
+      source: T.nilable(Proc),
+      key: T.nilable(T.any(String, Symbol)),
+      with_traits: T.nilable(T.any(Symbol, T::Array[Symbol])),
+      params: T::Hash[T.untyped, T.untyped],
+      options: T.untyped,
+      block: T.nilable(T.proc.bind(T.untyped).void),
+    ).void
+  end
+  def association(name, condition = nil, resource: nil, serializer: nil, source: nil, key: nil, with_traits: nil, params: T.unsafe(nil), **options, &block); end
+
+  sig do
+    params(
+      name: T.nilable(T.any(String, Symbol)),
+      if: T.nilable(T.any(Symbol, Proc)),
+      name_with_type: T.untyped,
+      block: T.proc.bind(T.untyped).params(object: T.untyped).returns(T.untyped),
+    ).void
+  end
+  def attribute(name = nil, if: nil, **name_with_type, &block); end
+
+  sig do
+    params(
+      attrs: T.any(String, Symbol),
+      if: T.nilable(T.any(Symbol, Proc)),
+      attrs_with_types: T.untyped,
+    ).void
+  end
+  def attributes(*attrs, if: nil, **attrs_with_types); end
+
+  sig { params(key: T.any(String, Symbol)).void }
+  def collection_key(key); end
+
+  sig do
+    params(
+      name: T.any(String, Symbol),
+      condition: T.nilable(Proc),
+      resource: T.nilable(T.any(Module, String, Proc)),
+      serializer: T.nilable(T.any(Module, String, Proc)),
+      source: T.nilable(Proc),
+      key: T.nilable(T.any(String, Symbol)),
+      with_traits: T.nilable(T.any(Symbol, T::Array[Symbol])),
+      params: T::Hash[T.untyped, T.untyped],
+      options: T.untyped,
+      block: T.nilable(T.proc.bind(T.untyped).void),
+    ).void
+  end
+  def has_many(name, condition = nil, resource: nil, serializer: nil, source: nil, key: nil, with_traits: nil, params: T.unsafe(nil), **options, &block); end
+
+  sig do
+    params(
+      name: T.any(String, Symbol),
+      condition: T.nilable(Proc),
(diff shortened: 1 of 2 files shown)


rbi/annotations/alba.rbi
```diff
@@ -0,0 +1,173 @@
+# typed: true
+
+module Alba::Resource::ClassMethods
+  sig do
+    params(
+      name: T.any(String, Symbol),
+      condition: T.nilable(Proc),
+      resource: T.nilable(T.any(Module, String, Proc)),
```
Morriar: I wonder if we should include `Symbol` here too, and in the copied association alias signatures below.

`Alba::Association#constantize` handles both `Symbol` and `String`, and `serializer:` goes through the same `resource ||= serializer` path. Without `Symbol`, typed code using `resource: :ArticleResource` or `serializer: :ArticleResource` will fail typechecking even though Alba supports it at runtime.

```suggestion
      resource: T.nilable(T.any(Module, String, Symbol, Proc)),
```

rbi/annotations/alba.rbi
```diff
@@ -0,0 +1,173 @@
...
+      options: T.untyped,
+      block: T.proc.bind(T.untyped).void,
+    ).void
+  end
+  def nested_attribute(name, **options, &block); end
+
+  sig do
+    params(
+      handler: T.nilable(Symbol),
+      block: T.nilable(T.proc.params(error: T.untyped, object: T.untyped, key: T.untyped, attribute: T.untyped, resource_class: T.untyped).returns(T::Array[T.untyped])),
```
KaanOzkan:
```suggestion
      block: T.nilable(T.proc.params(error: T.untyped, object: T.untyped, key: T.untyped, attribute: T.untyped, resource_class: T.untyped).returns(T.untyped)),
```

Looks like it can return any constant? https://github.com/okuramasafumi/alba/blob/v4.0.0/test/usecases/on_error_test.rb#L60-L63

rbi/annotations/alba.rbi
```diff
@@ -0,0 +1,173 @@
+# typed: true
+
+module Alba::Resource::ClassMethods
+  sig do
+    params(
+      name: T.any(String, Symbol),
+      condition: T.nilable(Proc),
+      resource: T.nilable(T.any(Module, String, Proc)),
+      serializer: T.nilable(T.any(Module, String, Proc)),
```
KaanOzkan: This seems better

```suggestion
      resource: T.nilable(T.any(T::Class[Alba::Resource], String, Proc)),
      serializer: T.nilable(T.any(T::Class[Alba::Resource], String, Proc)),
```

https://github.com/okuramasafumi/alba/blob/fd805a70f8bedbfd1cfc50eb589afa553469ace4/lib/alba/resource.rb#L392-L393

KaanOzkan: Similarly for `one`, `has_many`, `has_one`, `many`

rbi/annotations/alba.rbi
mokevnin: Good catch — the return goes straight back to the caller, so `T::Array[T.untyped]` was wrong. Applied in d63f627.

rbi/annotations/alba.rbi
mokevnin: Agreed, `T::Class[Alba::Resource]` is the honest type. Applied in d63f627.

mokevnin: Done — `association`, `one`, `many`, `has_one` and `has_many` all carry the same signature now. `bundle exec repo check --gem alba` is green.

KaanOzkan (approved):

PR #423, opened 2026-09-11
activesupport: bind ActiveSupport::Concern block DSLs to T.untyped

frewsxcv:
Adds annotations for `ActiveSupport::Concern#included`, `#prepended`, and `#class_methods` binding the block's `self` to `T.untyped`.

These methods `class_eval` their block on the including class (or the generated `ClassMethods` module), so `self` inside the block is never the concern module. Without a binding, Sorbet types `self` as `T.class_of(TheConcern)` and reports 7003 on every DSL call in an `included do … end` body (`scope`, `validates`, `has_many`, `class_attribute`, …), which pushes concern modules toward `# typed: false`.

`T.untyped` is the honest binding since a concern can't know its includer, and it matches the existing `Testing::Declarative#test` and `SetupAndTeardown#setup`/`#teardown` annotations in this file.

Version guards: `class_methods` since 4.2, `prepended` since 6.1.

`repo check index`, `rubocop`, `runtime activesupport`, and `static activesupport` all pass locally.


rbi/annotations/activesupport.rbi
```diff
@@ -5,6 +5,19 @@ module ActiveSupport::Testing::Declarative
...
+module ActiveSupport::Concern
+  sig { params(base: T.untyped, block: T.nilable(T.proc.bind(T.untyped).void)).void }
+  def included(base = nil, &block); end
+
+  # @version >= 6.1.0
+  sig { params(base: T.untyped, block: T.nilable(T.proc.bind(T.untyped).void)).void }
+  def prepended(base = nil, &block); end
+
+  # @version >= 4.2.0
+  sig { params(class_methods_module_definition: T.nilable(T.proc.bind(T.untyped).void)).void }
```
Morriar: Let's make this block required. `class_methods` evaluates the block with `module_eval`, and calling it without a block raises at runtime, so the annotation shouldn't typecheck `class_methods` with no block.

```suggestion
  sig { params(class_methods_module_definition: T.proc.bind(T.untyped).void).void }
```

amomchilov (approved):
Too bad that proc `bind`s can't be generic. `T.untyped` is the best we can do here, for now.  
  
Thanks!

frewsxcv:
@amomchilov Hmm is there something that needs to happen in order for the build in CI to get kicked off?

amomchilov:
@frewsxcv I forgot to push the button :)

Started.

PR #424, opened 2026-09-13
alba: bind class_eval'd DSL blocks to T.self_type

mokevnin:
### Type of Change

### Changes

* Gem name: alba
* Gem version: 4.0.0
* Gem source: https://github.com/okuramasafumi/alba/blob/main/lib/alba/nested_attribute.rb#L24-L27, https://github.com/okuramasafumi/alba/blob/main/lib/alba/resource.rb#L118-L122
* Gem API doc: https://rubydoc.info/gems/alba/Alba/Resource/ClassMethods
* Tapioca version: 0.19.2
* Sorbet version: 0.6.13497

Follow-up to #422, which is mine: I typed the blocks of `nested`, `nested_attribute` and `trait` as `T.proc.bind(T.untyped)`. That binding is weaker than the runtime warrants, and the erasure it causes is not local to the block's own `self`.

All three `class_eval` the block on a subclass of the resource class the macro was called on:

```ruby
# lib/alba/nested_attribute.rb
resource_class = Class.new(@klass)
resource_class.class_eval(&@block)

# lib/alba/resource.rb, #hash_from_traits
resource_class = Class.new(self.class)
resource_class.class_eval(&body)
```

So `self` inside the block is exactly the receiver of the macro, which `T.self_type` names precisely. This differs from the `ActiveSupport::Concern` case in #423: a concern genuinely cannot know its includer, whereas here the type is right there in the call.

The cost of `T.untyped` is that it also erases what is declared *inside* the block. A project that types its resources (a Tapioca DSL compiler generating a per-resource `attribute` sig, or a hand-written shim) loses that typing for everything inside `nested do ... end` — the block parameter, and every value derived from it, silently becomes untyped, because `attribute` is now being called on an untyped `self`. In our app this turned a correct `T.must(it.price)` on a `T.nilable(SubscriptionPrice)` into a 7050 "redundant T.must", i.e. the annotation did not just fail to add information, it removed information Sorbet previously had. With `T.self_type` that typing is preserved, and inside the block plain `nested`/`attribute` calls still resolve through `Alba::Resource::ClassMethods` as before.

`repo check index`, `rubocop alba`, `runtime alba` and `static alba` all pass locally.

Blocks left alone deliberately: `attribute`, `meta`, `on_nil` and the association `source` blocks are `instance_exec`'d on a resource *instance*, not the class, so `T.self_type` would be wrong there and the honest type is a separate question.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


KaanOzkan (approved):

mokevnin:
Thanks for the quick review and merge!

